PR #37, opened 2025-12-11
Fix SMB2 create problem on Linux/io_uring backends

benjarvis:
Add CHIMERA_VFS_OPEN_READ_ONLY flag, map NFS4 and SMB access attributes to it.   In SMB2 create, if client indicates it wants only to read attributes, specify CHIMERA_VFS_OPEN_PATH.  Add some more dump details to SMB2 create.

This is particularly important for linux/io_uring backends as the kernel cares and opening directories for read/write is not allowed.  For the other native backends, the VFS caller is presumed to be privileged, so this level of security is not necessary, consequently they ignore these flags.

This is not the whole story for properly handling all the SMB2 create options yet...

### DIFF
--- a/src/server/nfs/nfs4_proc_open.c
+++ b/src/server/nfs/nfs4_proc_open.c
@@ -149,7 +149,10 @@ chimera_nfs4_open_parent_complete(
             case EXCLUSIVE4_1:
                 break;
         } /* switch */
+    }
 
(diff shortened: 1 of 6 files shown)


benjarvis:
Ctest Failure here is unrelated to the change so I'll investigate that separately and merge it anyway.

PR #38, opened 2025-12-17
Add external_portmap option to server config.

arenaud16:
In some cases we want to rely on the system portmapper instead of deploying our own version this change
allow us to register to the system portmapper
if server option "external_portmap" is set to true


arenaud16:
Not sure how to add **libtirpc-dev** to the build container it is needed

PR #194, opened 2026-02-14
Add KVM-based Windows SMB integration tests

benjarvis:
## Summary

- Adds infrastructure to build a Windows Server 2025 QEMU image and run native PowerShell tests over mapped SMB shares against chimera's SMB server
- Exercises the real Windows SMB client stack (NTLM auth, file I/O, directory operations) rather than POSIX compatibility layers
- Three test cases per backend: `auth_ntlm`, `file_io`, `dir_ops`

## Known issues

- `file_io` test crashes chimera on memfs backend (ASAN abort in `chimera_smb_query_directory_readdir_complete`)
- `file_io` test fails on linux backend (`Remove-Item` reports file not found after successful write/read)
- These are pre-existing server bugs that the new tests correctly surface

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -140,6 +140,7 @@ else()
 endif()
 
 set(KVM_TESTING_AVAILABLE FALSE)
+set(KVM_WINDOWS_TESTING_AVAILABLE FALSE)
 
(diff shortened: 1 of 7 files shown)


benjarvis:
Not going to do this, instead aiming to ultimately add windows tests using native windows runners as there's no licensing concerns that way.

PR #255, opened 2026-05-19
S3: implement multipart upload with VFS-capability-driven assembly

benjarvis:
## Summary

Implements S3 multipart upload end-to-end, then makes the AWS-spec-deviant pieces of CompleteMultipartUpload / UploadPart / AbortMultipartUpload conformant.

### Commit 1 — phase 1 API surface
Wires up the six multipart operations: `CreateMultipartUpload`, `UploadPart`, `CompleteMultipartUpload`, `AbortMultipartUpload`, `ListParts`, `ListMultipartUploads`.

- In-memory multipart table on the shared S3 server state, RW-locked, with per-upload mutex + refcount so async `UploadPart` and `Abort` can interleave safely across worker threads.
- `UploadPart` writes each part into an unlinked file via `chimera_vfs_create_unlinked` when the destination VFS module sets `CAP_CREATE_UNLINKED`, otherwise a `._chimera_mpu_*` temp file — same fallback pattern as PUT.
- Bug fix in V4 canonical query string encoding: bare keys (`?uploads`) must be canonicalized as `uploads=` per the AWS spec; multipart auth was failing without this.

### Commit 2 — phase 2 part assembly
`CompleteMultipartUpload` walks the part list and stitches sources into the final object. The destination's VFS module capabilities select the primitive:

| Cap advertised | Path |
|---|---|
| `CAP_MOVE_RANGE` | `chimera_vfs_move_range` (memfs — zero-copy block transfer) |
| `CAP_COPY_RANGE` | `chimera_vfs_copy_range` (linux / io_uring — `copy_file_range(2)`) |
| neither | chunked `chimera_vfs_read` + `chimera_vfs_write` loop (cairn) |

Parts are held open through assembly; their file handles + tmp files are released by the existing async part-destroy chain after the final object is linked into place.

### Commit 3 — AWS compliance: manifest validation + part limits
Previous phases ignored the `CompleteMultipartUpload` body and always assembled the full server-side part list. Real S3 clients send a manifest specifying which parts (and in what order) to combine, and some legitimately complete with a subset. Now:

- Notifier accumulates the body for `POST`-with-`uploadId` instead of draining; new two-phase Complete flow waits for the body, parses it (hand-rolled XML scanner), validates against the server's recorded parts, and only then detaches the upload and starts assembly.
- Assembly walks the **client's** part list (subset and ordering honored), not the server's full list.
- New error paths: `InvalidPart` (ETag mismatch or part not found), `InvalidPartOrder` (manifest not strictly ascending), `EntityTooSmall` (non-final part < 5 MiB), `MalformedXML` (body unparseable or > 16 MiB).
- `UploadPart` rejects `PartNumber < 1 || > 10000` with `InvalidArgument`.
- `AbortMultipartUpload` returns HTTP 204 (was 200) via a new `NoContent` status mapping.
- Failed validation leaves the upload in the table for retry.

## Tests

`test_multipart` in `tests/s3_test.py` exercises the full lifecycle and asserts content correctness; registered against memfs / linux / io_uring / cairn so every assembly path runs the same scenarios:

- 3 equal 5 MiB parts — assembled object equals `part1 ‖ part2 ‖ part3`
- 3 varied-size parts (5 / 6 / 5 MiB)
- **Subset completion** — complete [1, 3] of 3 uploaded; result = `part1 ‖ part3`
- **Wrong ETag** → `InvalidPart`; verified upload survives for retry
- **Out-of-order manifest** → `InvalidPartOrder`
- **Nonexistent part_number** → `InvalidPart`
- **Two 1-MiB parts** → `EntityTooSmall`
- **`PartNumber=0` and `PartNumber=10001`** → `InvalidArgument`
- **Abort path** — uploaded part discarded, no final object materialized
- **`NoSuchUpload`** — `ListParts` on bogus upload id

Two backends are deliberately omitted from the multipart matrix with inline `CMakeLists.txt` notes:

- **demofs (V4)** — RW assembly works on cairn but corrupts data on demofs (read iov passed back to write on a different file). Suspected demofs bug, not in the S3 layer.
- **memfs (V2)** — V2 signing's canonical resource doesn't yet include the multipart subresources (`uploads`, `uploadId`, `partNumber`).

## Verification

- `make syntax-check` ✅
- `make build_release` ✅
- `make build_clang` (scan-build) ✅ — "No bugs found"
- `make reuse-lint` ✅
- `make test_debug`: **1894/1894 pass, 0 failed** (including all 34 V4 + V2 S3 tests)

## Known gaps not addressed in this PR

Documented for follow-up. Severity = interop impact for typical S3 client traffic.

**HIGH:**
- **ETag is XXH3-128, not MD5.** Real S3 per-part ETag = MD5 of part body, final = `MD5(concat-of-MD5s)-<count>`. Clients like aws-cli verify per-part ETags against locally-computed MD5; ours mismatch. (The manifest validation in this PR is unaffected — the client just echoes what we returned — but post-upload integrity checks will fail.)

**MEDIUM:**
- **No persistence.** Multipart state is in-memory only; lost on restart.
- **`CreateMultipartUpload` drops object metadata.** `Content-Type`, `Cache-Control`, `Content-Encoding`, `Content-Disposition`, `x-amz-meta-*` headers are not stored and don't apply to the final object.
- **No pagination in `ListParts`** (`max-parts`, `part-number-marker` ignored). Listing > ~1000 parts overflows the 1 MiB response buffer.
- **No pagination/filtering in `ListMultipartUploads`** (`max-uploads`, `key-marker`, `prefix`, `delimiter` ignored). No `CommonPrefixes`.

**LOW:**
- `UploadPartCopy` (PUT with `x-amz-copy-source` header) not implemented.
- No `Owner` / `Initiator` XML elements in ListParts / ListMultipartUploads.
- No SSE / SSE-C / SSE-KMS anywhere in chimera S3 (pre-existing).
- No `HEAD` on individual parts.
- No keep-alive newlines during long `CompleteMultipartUpload` assembly.
- V2 signing canonical resource doesn't include multipart subresources.
- `CompleteMultipartUpload` returns HTTP 400 on validation failure, not AWS's "200 with error body" pattern.


benjarvis:
This is not a done deal yet but given we have nothing now I'll merge it and address its shortcomings in future PRs.

PR #259, opened 2026-05-19
smb: tighten SMB2 READ validation and track per-handle position

benjarvis:
## Summary

Bring `chimera_smb_read` in line with MS-SMB2 §3.3.5.12 for protocol corners exercised by Samba's `smb2.read` smbtorture suite:

- **`INVALID_DEVICE_REQUEST`** when reading a directory handle
- **`ACCESS_DENIED`** when the open lacks `FILE_READ_DATA`/`FILE_EXECUTE`
- **`INVALID_PARAMETER`** for negative offsets, `offset + length` overflow, `Length > MaxReadSize`, or `MinimumCount > Length`
- **`END_OF_FILE`** when actual bytes read are below `MinimumCount`

Also advances `open_file->position` by the number of bytes read so that subsequent `FileAllInformation` queries see a correct file position (previously stuck at 0 unless the client explicitly set it via `SET_INFO POSITION_INFO`).

Unblocks 4 of 5 `smb2.read` subtests under smbtorture (`eof`, `position`, `dir`, `access`). The remaining one (`bug14607`) needs a samba-test-only FSCTL.

### DIFF
--- a/src/server/smb/smb_proc_read.c
+++ b/src/server/smb/smb_proc_read.c
@@ -25,6 +25,10 @@ chimera_smb_read_callback(
     int                               chunk_niov;
     int                               i;
 
+    if (!error_code) {
+        request->read.open_file->position = request->read.offset + count;
+    }
+
     chimera_smb_open_file_release(request, request->read.open_file);
 
     request->read.niov     = niov;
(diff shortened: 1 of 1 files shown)


benjarvis:
Pushed a follow-up commit (c534821) to fix the SMB CI failures.

The original commit's strict `FILE_READ_DATA | FILE_EXECUTE` check rejected the Linux CIFS client, which opens files with `GENERIC_READ` (0x80000000) and never the specific `FILE_READ_DATA` bit. That caused ~50 cthon/xfstests SMB tests to fail with `STATUS_ACCESS_DENIED` across all backends. The follow-up broadens the check to also accept the generic and maximum-allowed bits — `smbtorture smb2.read.access` still hits the deny path because it deliberately opens with none of these set.

PR #273, opened 2026-05-19
Unified lease/lock backend (NFSv3, NFSv4, SMB2 share/range/oplock)

benjarvis:
## Summary

Introduces a single VFS-layer state object (\`vfs_state\`) that coordinates byte-range locks, share-mode reservations, and breakable caching leases across NFSv3 (NLM), NFSv4, and SMB2. The protocol layers stop maintaining their own per-protocol lock tables and instead acquire/release leases through one shared backend with a unified conflict matrix.

The motivation: today NFS and SMB lock state are completely disjoint, so an NFS write doesn't block a conflicting SMB open and vice versa. Once Chimera grants SMB leases or NFSv4 delegations, that incoherence becomes a correctness problem — the holder caches based on assumptions the other protocol can't honor. \`vfs_state\` is the single source of truth all three protocols consult.

## Design at a glance

Three lease kinds on one per-file state object:

- \`RANGE\` — byte-range locks (NLM, NFSv4 LOCK, SMB2 LOCK)
- \`SHARE\` — whole-file share/deny reservations (NFSv4 OPEN, SMB CREATE \`share_access\`)
- \`CACHING\` — breakable caching leases (NFSv4 delegations, SMB2 leases / oplocks)

Conflict matrix lives in one place. Same-owner coalescing implements the SMB lease-key rule (multi-open by one client shares one lease) and the NFSv4 lock-owner rule. SMB2 RWH bits are the lingua franca; per-protocol mappers translate at call sites.

Async \`chimera_vfs_lease_acquire\` supports waiting on breakable conflicts via a per-file pending queue; break orchestration invokes the holder's \`break_cb\`, transitions IDLE → BREAKING → ACKED/REVOKED, and pumps the queue so waiting acquires retry.

## What's in scope here

- **Foundation** (\`src/vfs/vfs_state.{h,c}\` + 17 unit tests)
- **Range locks**: NFSv4 LOCK/LOCKT/LOCKU migrated; LOCKU now releases the precise lease (previously fcntl unlock without tracking); NLM TEST/LOCK/UNLOCK migrated, per-host conflict scan deleted; NLM CANCEL implemented (was a stub returning GRANTED); new SMB2_LOCK opcode dispatch with parse/handler/reply.
- **Share modes**: SMB CREATE share-mode check routes through vfs_state (legacy sharemode table bypassed); NFSv4 OPEN \`share_access\` / \`share_deny\` honored (previously a single-flag no-op at line 298).
- **SMB caching leases**: CREATE grants SMB2 leases (RqLs v1/v2) and legacy oplocks; new \`smb_proc_oplock_break.c\` builds and sends the OPLOCK_BREAK Notification on the conn that owns the lease; SMB2_OPLOCK_BREAK ack handler echoes the Response packet; \`chimera_smb_conn_free\` clears \`create_conn\` so \`break_cb\` never dereferences a freed bind.

## Behavior changes worth flagging

- NFSv4 OPEN can now return \`NFS4ERR_SHARE_DENIED\` (previously always granted regardless of share_deny).
- NFSv4 LOCKU returns \`NFS4ERR_LOCK_RANGE\` for unmatched ranges (previously called fcntl unlock blindly).
- SMB2 byte-range LOCK / UNLOCK now actually work (previously \`STATUS_NOT_IMPLEMENTED\`).
- SMB CREATE can now grant SMB2 leases and oplocks (previously always \`OPLOCK_LEVEL_NONE\`).
- **Regression**: NFSv3 byte-range locks no longer survive a server restart — the NLM on-disk lease list is removed per the approved in-memory-only plan. The 90s grace window still opens; it just finds nothing to reclaim.

## What's deferred

- **NFSv4 callback channel + delegations**. OPEN still returns \`OPEN_DELEGATE_NONE\`. Granting delegations requires a server-initiated RPC direction (CB_COMPOUND / CB_RECALL) which doesn't exist in the codebase yet — needs new infrastructure in \`evpl_rpc2\` plus BIND_CONN_TO_SESSION, DELEGRETURN, RENEW. Slated for a follow-up.
- **Persistence**. All vfs_state lives in memory. Reclaim handshakes (NLM grace already wired; NFSv4 \`CLAIM_PREVIOUS\`+\`reclaim=1\`; SMB3 DH2Q/DH2C) work across client disconnect within one server lifetime but not across crashes. Persistence is a separate follow-up (per-backend journaling vs outboard KV — undecided).

## Verification

- 1891 tests pass debug + release. NLM lockf and cthon tlock suites exercise the migrated NLM path; \`vfs_state_test.c\` exercises the full conflict matrix in isolation.
- \`scan-build\` clean (debug + release).
- \`reuse-lint\` clean, \`syntax-check\` clean.
- The libevpl network tests (\`bulk_msg_tcp\`, \`ping_pong_msg_tcp\`, \`bulk_stream_tcp\`, \`ping_pong_msg_tls\`) are intermittently flaky under parallel load and unrelated to this change; all pass when re-run solo.

## Commit layout

1. \`vfs: add unified lease/lock state backend\` — additive, no behavior change, 17-test suite included.
2. \`server: route NLM, NFSv4, and SMB locking through vfs_state\` — wires every protocol through the new backend.

### DIFF
--- a/src/server/nfs/nfs.c
+++ b/src/server/nfs/nfs.c
@@ -519,7 +519,9 @@ chimera_nfs_server_notify(
                 if (nlm_cli->conn_count == 0) {
                     /* Last connection for this hostname -- release locks */
                     nlm_client_release_all_locks(&shared->nlm_state, nlm_cli,
-                                                 thread->vfs_thread, &anon_cred);
+                                                 thread->vfs_thread,
+                                                 thread->vfs->vfs_state,
+                                                 &anon_cred);
                     release_locks = true;
                 }
                 pthread_mutex_unlock(&shared->nlm_state.mutex);
--- a/src/server/nfs/nfs4_proc_lock.c
+++ b/src/server/nfs/nfs4_proc_lock.c
@@ -2,12 +2,16 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-only
 
+#include <stdlib.h>
+#include <xxhash.h>
+
 #include "nfs4_procs.h"
 #include "nfs4_status.h"
 #include "nfs4_session.h"
 #include "nfs4_state.h"
 #include "vfs/vfs_procs.h"
 #include "vfs/vfs_release.h"
+#include "vfs/vfs_state.h"
 
 /*
  * RFC 7530 §9.1.7 LOCK completion wrapper.  Advances the owner seqid(s)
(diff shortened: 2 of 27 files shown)


benjarvis:
Rebased onto current `main` (picks up #253's unified NFSv4 state-model rewrite and the libevpl bump).

Since #253 replaced the slot-array `nfs4_state` model this branch was originally built on, the NFSv4 lock/share integration was reimplemented against the new `nfs_open_state` / `nfs_lock_state` / `nfs_state_table` model (third commit). Key points:

- NFSv4 LOCK/LOCKU/LOCKT now route through `vfs_state` instead of the legacy `chimera_vfs_lock` backend. Per-`nfs_lock_state` range-lease list; LOCKU/LOCKT are now synchronous.
- NFSv4 OPEN acquires a `vfs_state` SHARE lease in `chimera_nfs4_open_install_state`, layering cross-protocol (NLM/SMB) coordination on top of upstream's intra-client `nfs_client_check_share_conflict`.
- Lease release is hooked into `open_state_cleanup` / `lock_state_cleanup` — the refcounted teardown choke points — so CLOSE, lock-owner teardown, and client expiry all release correctly.
- Side benefit: NFSv4 byte-range locks now work on all backends (memfs/cairn/demofs), not just fcntl-backed ones.

NLM and SMB integration is unchanged from the original push. NFSv4 delegations + the callback channel remain deferred (unchanged). All 2396 tests pass debug + release; scan-build clean.

PR #347, opened 2026-05-22
diskfs: persistent, crash-consistent on-disk demofs (b+tree, intent log, caches, zero-copy/COW, space reclaim)

benjarvis:
This branch turns `demofs` into a persistent, crash-consistent on-disk filesystem — an async, journaled VFS backend — built up over the commits on `diskfs`.

## What's here

**On-disk structure + async engine**
- Per-inode on-disk b+tree (dirents, extents, symlink targets) with split/merge/join, driven by a fully async op machinery (block waiter + doorbell + resumable op driver). Every VFS entry point (lookup, readdir, read, write/RMW, setattr/truncate, seek, mkdir/mknod/open/symlink/link/remove/rename) is converted to async b+tree ops.
- Log-structured AG space allocator (condense + reconstruct); per-transaction allocator deltas are journaled through the redo log (crash-safe allocator).

**Durability / crash recovery**
- Write-ahead intent log (full-block redo) with XXH3-128 per-record checksums + embedded tail; synchronous replay on mount; real superblock with CRC + clean/unclean detection. File data is written to its home location before the metadata transaction commits, and the inode write lock is held until the record is durable (dirty blocks are never visible to another thread before the log protects them).
- Tail-pusher pushes committed block images home and trims the log.

**Caches + eviction**
- Block cache: 256 shards, fixed pre-allocated pool with LRU recycling, provisioned above the journal so it never blocks (asserts on exhaustion rather than deadlocking).
- Inode cache: per-shard LRU eviction of idle inodes (recycle + re-fault), with async flag-based inode locks that suspend the operation rather than block the worker.

**Zero-copy hot path + COW**
- Block buffers are SHARED evpl iovecs; the intent logger and tail-pusher reference them zero-copy — a block image is copied only when a writer forks a still-referenced LOGGED block (COW) — with a streaming-checksummed scatter-gather on-log record format.

**Metadata space reclaim**
- Pending free-on-commit (a freed range's delta is journaled, but the in-memory free is withheld until the txn is durable). b+tree shrink reclaims merged node blocks; small deleted inodes reclaim inline; large deleted inodes are reclaimed by a background incremental drainer (bounded batches per transaction) backed by a durable orphan-list inode for crash-safe resume.

## Validation
fsx (truncate / eviction / COW churn) and the demofs ctest suite pass under ASan, modulo the pre-existing NFS4-layer flakes noted below; remount/recovery is clean.

## Known follow-ups (intended for parallel work)
- **Durability barrier — highest priority.** Writes currently complete from the device's volatile cache: there is no fsync/FUA/barrier (devices are O_DIRECT and the `sync` arg to the block layer is ignored). So diskfs is crash-*consistent* against process crashes (kill -9, daemon restart) but **not durable against power loss**. Needs a flush/FUA at journal commit and before log trim, with group commit.
- I/O errors currently abort the daemon instead of returning EIO and staying mountable.
- Transaction abort does not roll back in-memory allocator *alloc* deltas (frees are already pending-on-commit).
- Incremental-drainer crash-resume (the orphans-found branch) is built on validated pieces but lacks a persistent-device crash-injection test harness; the `DEMOFS_DRAIN_DISABLE` knob is in place to drive one.
- Deleted-inode in-memory struct leak (the space is reclaimed; the struct is not); block-buffer fragmentation (evpl 2 MiB sub-allocation vs 4 KiB blocks); inode-cache hard-cap waiter.
- Pre-existing NFS4-layer flakes (`cthon`/`nametest`/`rewinddir` on `*_nfs4_demofs_*`) — confirmed independent of this work.

### DIFF
--- a/src/vfs/demofs/CMakeLists.txt
+++ b/src/vfs/demofs/CMakeLists.txt
@@ -1,9 +1,10 @@
-# SPDX-FileCopyrightText: 2025 Chimera-NAS Project Contributors
+# SPDX-FileCopyrightText: 2025-2026 Chimera-NAS Project Contributors
 #
 # SPDX-License-Identifier: LGPL-2.1-only
 
 add_library(chimera_vfs_demofs SHARED
     demofs.c
+    space_map.c
 )
 
 target_link_libraries(chimera_vfs_demofs jansson)
--- a/src/vfs/demofs/demofs.c
+++ b/src/vfs/demofs/demofs.c
@@ -1336,8 +1336,6 @@ demofs_block_claim(
     struct demofs_block_shard *shard  = &cache->shards[sidx];
     struct demofs_block       *blk;
 
-    (void) is_new;
-
     pthread_mutex_lock(&shard->lock);
 
     blk = demofs_block_lookup_locked(shard, bucket, device_id, device_offset);
@@ -1348,6 +1346,28 @@ demofs_block_claim(
         blk->buffer        = calloc(1, DEMOFS_BLOCK_SIZE);
         blk->state         = DEMOFS_BLOCK_CLEAN;
         blk->pin_count     = 0;
+
+        /* On a remounted FS a synchronous structural modify can reference a
+         * block the async descent never faulted in -- notably a leaf merge
+         * relinking the right partner's next_leaf, which is not one of the
+         * ci-1/ci/ci+1 siblings the rebalance pre-faults.  Read its real
+         * contents from disk rather than publishing a zeroed block (which
+         * would corrupt the tree once written back).  Rare; a buffered read
+         * served from the page cache.  is_new blocks are freshly allocated,
+         * so they are correctly left zeroed. */
+        if (thread->shared->mounted && !is_new) {
+            int fd = open(thread->shared->device_paths[device_id], O_RDONLY);
+
+            if (fd >= 0) {
+                ssize_t n = pread(fd, blk->buffer, DEMOFS_BLOCK_SIZE,
+                                  (off_t) device_offset);
+                close(fd);
+                chimera_demofs_abort_if(n != (ssize_t) DEMOFS_BLOCK_SIZE,
+                                        "block_claim disk read failed off=%lu n=%zd",
+                                        device_offset, n);
+            }
+        }
+
         /* Publish into the bucket chain (RCU readers will use hash_next). */
         blk->hash_next = shard->buckets[bucket];
         rcu_assign_pointer(shard->buckets[bucket], blk);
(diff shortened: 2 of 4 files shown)


benjarvis:
Pushed `3f1c2d2` fixing two intent-log-thread crashes that the heavier CI parallelism exposed (both around its redo + tail-push block queue under load; neither reproduced in isolation):

1. **`too many pending iocbs`** — `demofs_iq_process_channel` drained every channel's whole submission queue per wake with no cap on the block backend's submission ring (depth 256). A burst of queued txns overran it before completions drained it. Now bounded by an in-flight block-write counter (`DEMOFS_IL_IOCB_CAP`) with a low-watermark resume from write completions.

2. **`failed to write to doorbell fd`** — `demofs_intent_log_thread_shutdown` removed `wake_doorbell` *before* the shutdown drain loop, but a tail-push completion in that loop (`demofs_il_push_finish`) rings it to re-poke channel processing → write to a closed eventfd → abort. Only triggers when the tail-pusher is behind at unmount, hence load-dependent. Now drains first, then drops the doorbell.

Validation: a `-j24` repro of the io_uring/aio cthon + rewinddir demofs tests crashed on iteration 1 before; now 40/40 clean. Full demofs posix subset 4/4 clean at `-j16`. Debug+Release build clean, `make syntax` clean.

benjarvis:
Pushed three more fixes (`c6f29af`, `869ac9a`, `0d21141`) for bugs the **KVM xfstests/fsstress/SMB suite** exposed — these jobs (`Test devcontainer linux/amd64`) only started running once the build was fixed, and surfaced 4 *pre-existing* diskfs bugs (not regressions):

1. **`b+tree record too large`** (`c6f29af`) — `op->recbuf` is sized for a dirent name (255B), but the generic b+tree insert also stores a **symlink target**; longer targets aborted the daemon. Stage oversized records on the heap; reject targets larger than one node with `ENAMETOOLONG`.

2. **`too many pending iocbs` / `io_uring_get_sqe`** (`869ac9a`) — no admission control on worker data I/O (fragmented reads fan out per-extent × `nconnect` concurrency) nor on the intent-log tail-push (issued a whole record's home writes at once). Added a per-thread in-flight cap that parks and resumes requests from block-I/O completions, plus chunked tail-push sharing the redo iocb budget. (Every `pending_io` decrement resumes parked requests — including the metadata-read completion, which otherwise deadlocks.)

3. **space_map `double-free or overlap`** (`0d21141`) — DEALLOCATE/zero-range/RMW-trim reinserts extents at sub-block (unaligned) device offsets (the read path supports this via padding); freeing them released a block a neighbouring extent still mapped. Now free only the whole blocks strictly contained in the range, consistently in the journal + apply halves.

Validation: all 18 `ubuntu2404` KVM demofs tests (fsstress + fsx × nfs3/4.0/4.1/4.2, both io_uring & aio, + SMB special cthon) pass, 0 daemon aborts; `fsx -N 5000` nfs4.2 nconnect=16 completes A-OK where it previously aborted. Debug+Release+`make syntax` clean.

PR #530, opened 2026-05-26
diskfs: avoid reusing cached deleted inodes

benjarvis:
## Summary
- keep deleted inode home blocks allocated while inode structs remain cached
- prevent new creates from reusing an inum that is still present in the inode cache
- update orphan-drain comments to match the current leak-until-generation-safe behavior

## Testing
- make syntax
- git diff --check upstream/main...HEAD
- cmake --build /tmp/diskfs-fail-build --target chimera_vfs_diskfs -j 8


benjarvis:
Superseded by #531, which includes this deleted-inode reuse fix plus the diskfs metrics/RMW changes.

Issue #79, opened 2026-01-15
Fix FSSTAT abytes field - should be available bytes, not used bytes


benjarvis:
## Summary

The FSSTAT procedure incorrectly populates the `abytes` field with "space used" instead of "available bytes".

## Current State

In `src/server/nfs/nfs3_proc_fsstat.c:36`:
```c
res.resok.abytes = attr->va_fs_space_used;  // WRONG
```

## Required Functionality

Per RFC 1813:
- `tbytes`: Total bytes on filesystem
- `fbytes`: Free bytes on filesystem
- `abytes`: Free bytes available to requesting user (may be less than fbytes due to quotas/reserved space)

## Fix

```c
res.resok.abytes = attr->va_fs_space_avail;  // Or similar available-space field
```

If VFS doesn't have an "available to user" field, `abytes` should equal `fbytes` as a reasonable default.

## Priority

Medium - Incorrect values may cause client-side space calculations to fail.

PR #547, opened 2026-05-27
submodule: bump libevpl so a malformed peer reply no longer aborts the server

benjarvis:
## Summary

A peer sending an RPC reply that fails to decode (truncated, malformed, or `len != length`) made libevpl's generated reply dispatcher return an error that `rpc2.c` turned into `evpl_rpc2_abort()` — **crashing the entire chimera process**. A client (or a buggy/dying peer) could take the server down.

This bumps `ext/libevpl` to pull in the rpc2 hardening: such a reply is now delivered to the request's completion callback as a failed call (`EVPL_RPC2_REPLY_DECODE_ERROR`, reply `NULL`) instead of aborting, so the caller cleans up and the server stays up. chimera's rpc2 client callbacks already check `status` before touching the reply.

## How it was found

The NFSv4.1 delegation callback path: when a client's callback server dies mid-reply and returns a truncated `CB_COMPOUND` reply (pynfs DSESS9003 does exactly this), chimera aborted with `evpl_rpc2_abort("Failed to dispatch rpc2 reply: 2")`. With this bump, chimera instead completes the recall with an error (revoking the delegation / marking the callback path down) and stays running.

## Dependency chain

- chimera-nas/xdrzcc#8 — codegen change.
- chimera-nas/libevpl#79 — adds `EVPL_RPC2_REPLY_DECODE_ERROR` + bumps `ext/xdrzcc` to #8.
- This PR — bumps `ext/libevpl` to #79.

Merge order: xdrzcc#8 → libevpl#79 (re-point `ext/xdrzcc` to the merged commit) → this PR (re-point `ext/libevpl` to the merged commit).

## Testing

- All 1590 `posix/*_nfs3*` + `posix/*_nfs4*` client tests pass (these exercise the regenerated reply dispatcher heavily) — no regression in normal reply handling.
- Under the Debug/AddressSanitizer build, the malformed-`CB_COMPOUND` scenario that previously aborted the server now completes cleanly and the daemon shuts down normally.

Note: this fixes the server-crash facet only; pynfs DSESS9003 itself still cannot pass (its py3.14 callback handler kills its own poll thread on the deliberately-unanswered recall, independent of the server) and remains disabled.


benjarvis:
Superseded by #551, which bumps ext/libevpl to the now-merged libevpl#79 (the malformed-reply server-crash fix) and builds the DSESS9003 recall-retransmit feature on top. #551 points at the merged libevpl commit rather than this PR's pre-merge branch pointer.

PR #585, opened 2026-05-29
smb2: generic async-interim (STATUS_PENDING) response infrastructure

benjarvis:
## Summary

Adds a reusable defer-and-emit-interim path that the dispatch site arms
on defer-eligible commands (READ, WRITE, FLUSH) when they are the tail
of their compound:

- Handler completes within 500us window -> `chimera_smb_complete_request`
  cancels the timer, request replies synchronously, no interim sent.
- Underlying VFS op takes longer -> timer fires a standalone SMB2
  `STATUS_PENDING` with `SMB2_FLAGS_ASYNC_COMMAND` and an `AsyncId` set to
  the original MessageId (Samba's convention).  The eventual final reply
  picks up `request->async_id != 0` and tags itself with the same flag.

New file `src/server/smb/smb_async_interim.c` hosts the
`arm` / `cancel` / `fire` / `drain` helpers.  `arm` snapshots the
signing key, dialect, session id, credit fields, and signed-session bit
on the request itself -- same pattern `smb_proc_change_notify.c` uses --
so the fire path does not race with concurrent session teardown for the
signing material.

## State threading

- `chimera_smb_request.async` substruct (timer + park_next + signing
  snapshot + armed) reuses the existing free-list-allocated request, no
  extra alloc cost; `chimera_smb_request_alloc` zeros it.
- `chimera_smb_conn.parked_requests` singly-linked list head; CANCEL
  walks it by AsyncId; `chimera_smb_conn_free` drains it before tearing
  down the bind.
- `chimera_smb_complete_request` gains a one-line cancel hook at the
  top.  `chimera_smb_compound_advance` arms in front of the dispatch
  switch when the command is defer-eligible AND this slot is the tail.

The whole machinery runs on the request's owning conn thread (dispatch,
timer fire, complete_request all serialise on the single-threaded evpl
loop), so no locks are needed.

## Mid-chain async deliberately NOT failed with INTERNAL_ERROR

MS-SMB2 reserves `STATUS_INTERNAL_ERROR` for a mid-compound op that goes
async, but chimera's current READ/WRITE/FLUSH producers either complete
in-line or via a callback that marshals back to the conn thread, so no
interim ever fires mid-chain today.  A mid-chain ban would falsely fail
a legitimately-fast mid-compound `FLUSH+WRITE`, `WRITE+CLOSE`, etc.
When a future producer (oplock break, lease break, blocking lock) adds
a genuinely long-running mid-chain op, the rule can be added alongside
it.

## CHANGE_NOTIFY left bespoke

`smb_notify.c`'s existing interim path is correct and already tested.
The two paths do not conflict (separate `parked_notifies` /
`parked_requests` lists; CANCEL walks both).  Folding notify in is a
later cleanup, not gated on test wins.

## Test wins

**None by itself.**  The infrastructure only ever emits an interim when
an underlying op exceeds 500us, which none of chimera's current VFS
operations do.  This is foundation for the eventual oplock-break /
lease-break / blocking-lock work.

## Validation (no regressions)

- `smb2.compound_async` stays 4/10 (Phase 1 wins from #584 intact).
- `smb2.compound` 9/20.
- `smb2.notify-inotify` green.
- Memfs sweep: compound, compound_find, compound_async, read, rw,
  sharemode, check-sharemode, create_no_streams, connect, tcon, mkdir,
  winattr2, secleak, timestamps, notify-inotify.
- Cairn + io_uring cross-backend: 9 compound + 4 compound_async on each.

## Stack

Stacked on **#584** (`smb-compound-fileid-propagate`), which itself stacks
on **#583** (`smb-compound-validation`).

### DIFF
--- a/src/server/smb/CMakeLists.txt
+++ b/src/server/smb/CMakeLists.txt
@@ -12,7 +12,7 @@ add_library(chimera_smb SHARED
     smb_proc_reparse.c smb_proc_change_notify.c smb_proc_cancel.c
     smb_proc_sparse.c smb_proc_copychunk.c
     smb_proc_lock.c smb_proc_oplock_break.c
-    smb_notify.c smb_sharemode.c smb_ntlm.c smb_durable.c
+    smb_notify.c smb_async_interim.c smb_sharemode.c smb_ntlm.c smb_durable.c
     smb_wbclient.c smb_auth.c smb_gssapi.c
     #idl/ndr_dcerpc.c
 )
--- a/src/server/smb/smb.c
+++ b/src/server/smb/smb.c
@@ -20,6 +20,7 @@
 #include "server/server.h"
 #include "evpl/evpl.h"
 #include "smb_internal.h"
+#include "smb_async_interim.h"
 #include "smb_wbclient.h"
 #include "smb_procs.h"
 #include "smb_notify.h"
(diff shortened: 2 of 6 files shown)


benjarvis:
Rehashing this into some related OpLock work that will land in a different PR

PR #605, opened 2026-05-31
nfs/client: RDMA zero-copy reads land in the caller's buffer (read_into Phase 2)

benjarvis:
## What

Phase 2 of `read_into` — the RDMA zero-copy payoff. When a read carries caller-provided destination buffers (`chimera_vfs_read_into`, from #603) and the upstream NFS connection is RDMA, the NFS client arms the READ's RPC **write chunk** to *be* the caller's buffer (`evpl_rpc2_conn_set_write_chunk_dest`), so the server RDMA-writes the reply data straight into it — **no copy** — and sets `read.landed_in_dest` so the VFS core skips its scatter-copy fallback.

- `nfs3_read.c` (NFS3 READ) and `nfs4_read.c` (NFS4 READ COMPOUND) both arm it. The write chunk carries the READ data field exactly as the internally-allocated chunk did; only the buffer source changes.
- Single-iovec dest only; **TCP and multi-iovec dest fall back to the Phase-1 copy** (correct, just not zero-copy).

End to end: `fio` (NFS3-RDMA) → `chimera_read_into` (#603) → VFS core → NFS client arms the write chunk → server RDMA-writes into fio's `xfer_buf`. No per-read copy anywhere on the read path.

## Submodule

Bumps libevpl to the `rpc2-write-chunk-dest` branch tip for `evpl_rpc2_conn_set_write_chunk_dest()`. **Re-bump to libevpl `main` after chimera-nas/libevpl#85 merges.**

## Dependencies / stacking

- Depends on **#603** (read_into Phase 1) and **chimera-nas/libevpl#85**.
- Stacked on the #603 branch — until #603 merges, this PR's diff also shows the Phase-1 commit. It shrinks to just the NFS changes + submodule bump once #603 lands.

## Validation

libevpl#85 adds a `rdma_ddp` `read_into` case proving the server RDMA-writes into the exact armed buffer (zero-copy) over the software-RDMA emulation. The chimera RDMA path itself needs a live RDMA rig — author validation on node-3: the `fio` NFS3-RDMA read job should show correct data (`verify=crc32c`) with the old per-read copy gone from `perf`, and throughput/CPU improved vs the Phase-1 copy path. Builds clean; Phase-1 memfs copy path (the `client_read_into` test) still green.

### DIFF
--- a/src/client/CMakeLists.txt
+++ b/src/client/CMakeLists.txt
@@ -4,7 +4,7 @@
 
 add_library(chimera_client SHARED client.c client_dispatch.c
             client_mount.c client_umount.c
-            client_open.c client_mkdir.c client_mknod.c client_close.c client_read.c client_write.c
+            client_open.c client_mkdir.c client_mknod.c client_close.c client_read.c client_read_into.c client_write.c
             client_symlink.c client_link.c client_remove.c client_rename.c
             client_readlink.c client_stat.c client_readdir.c client_statfs.c
             client_copy_range.c client_clone_range.c)
--- a/src/client/client.h
+++ b/src/client/client.h
@@ -201,6 +201,29 @@ chimera_writerv(
     chimera_write_callback_t        callback,
     void                           *private_data);
 
+/* Zero-copy read into caller-provided evpl_iovec(s).  Like chimera_read(), but
+ * the data lands directly in `iov` (over RDMA the buffers become the server's
+ * write target -- no copy).  The caller owns `iov` (borrow): keep the buffers
+ * alive until the callback fires, then release them.  The callback reports the
+ * byte count and eof; the data is already in the caller's buffers. */
+typedef void (*chimera_read_into_callback_t)(
+    struct chimera_client_thread *thread,
+    enum chimera_vfs_error        status,
+    uint32_t                      count,
+    uint32_t                      eof,
+    void                         *private_data);
+
+void
+chimera_read_into(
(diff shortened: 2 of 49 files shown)


benjarvis:
**Reworked the libevpl integration to use an explicit API** (code-therapy pass). The earlier approach passed the RDMA write-chunk destination through a one-shot connection side-channel (`evpl_rpc2_conn_set_write_chunk_dest`) to avoid changing the RPC call signature. That's now replaced with an explicit, optional `send_call_*` argument — clearer intent, no hidden per-connection state.

What changed here:
- `nfs3_read`/`nfs4_read` pass the read_into destination buffer directly through the new `write_chunk_iov`/`write_chunk_niov` arguments (still set `landed_in_dest` so the VFS core skips its copy).
- Every NFS3/NFS4/NLM/portmap/callback `send_call_*` site (~70) threads `NULL`/`0` through the new arguments to match the regenerated signature.
- Submodule bumped to libevpl `rpc2-write-chunk-param` (chimera-nas/libevpl#87), which carries the xdrzcc codegen change (chimera-nas/xdrzcc#9).

Validated: clean build, all 7 libevpl rpc2 tests + the chimera `client_read_into` test + 12 NFS3 client tests (symlink/link/remove/rename/readlink/stat) pass.

**Dependency chain now:** xdrzcc#9 → libevpl#87 → this PR. Re-bump each submodule to its repo's `main` as the PRs merge up the chain.

PR #621, opened 2026-06-02
vfs/smb: round-trip caller-supplied timestamps on setattr (smb2.setinfo)

benjarvis:
MS-FSCC 2.4.7 `SetFileInformation(FileBasicInformation)` lets a client set the file's `CreationTime` / `LastAccessTime` / `LastWriteTime` / `ChangeTime` explicitly, or omit any of them by sending 0 (or the freeze/thaw sentinels) to mean "don't change this field". In particular:

    SetInfo(BASIC, ctime = T)                  -> ChangeTime = T
    SetInfo(BASIC, ctime = 0, attrib = NORMAL) -> ChangeTime = T (preserved)

The smb2.setinfo conformance suite verifies this: a zero ChangeTime in a FileBasicInformation SetInfo MUST NOT bump the stored ChangeTime, even though the co-present DOS-attribute change is a metadata write that would otherwise advance ctime under POSIX.

### Two bugs blocking this

1. **`chimera_smb_unmarshal_basic_info`** only set `CHIMERA_VFS_ATTR_{ATIME,MTIME,CTIME,BTIME}` in the request's `set_mask` when the wire field was non-zero. An omitted ChangeTime therefore left `CHIMERA_VFS_ATTR_CTIME` out of `set_mask` entirely, and `apply_attrs` (below) bumped ctime to "now" because it could not tell the caller meant "preserve".
2. **`memfs_apply_attrs` / `cairn_apply_attrs` / `diskfs_apply_attrs`** unconditionally stamped `inode->ctime = now` at the end of every setattr, ignoring any explicit ChangeTime the caller supplied — so even `SetInfo(BASIC, ctime = T)` was overwritten with the current wall clock.

### Fix

- New `CHIMERA_VFS_TIME_OMIT` sentinel ("preserve the stored value, don't bump implicitly"), peer to the existing `CHIMERA_VFS_TIME_NOW`.
- `chimera_smb_unmarshal_basic_info` now always carries `ATIME/MTIME/CTIME/BTIME` in the request `set_mask`, with `TIME_OMIT` in any field the client zeroed. That's the explicit signal `apply_attrs` needs to suppress the implicit ctime bump when the co-present change is just a DOS-attribute write.
- `memfs` / `cairn` / `diskfs` `apply_attrs` honour `TIME_OMIT` for all four timestamps (skip the write) and honour an explicit `CHIMERA_VFS_ATTR_CTIME` in `set_mask`: `TIME_NOW` → bump to now, `TIME_OMIT` → preserve, any other value → store as-is. **ctime still bumps implicitly when CTIME is *not* in `set_mask`** (NFS chmod/chown, S3 metadata change), preserving POSIX semantics for non-SMB callers.
- `linux` / `io_uring` `set_attrs` map `TIME_OMIT` to `UTIME_OMIT` and skip the `utimensat` call entirely when no atime/mtime would actually change (avoids a spurious syscall on a pure attribute-bit SetInfo from SMB).

### What this greens

- **smb2.setinfo**'s `change_time round-trip` check now passes on memfs/cairn/diskfs. The suite still trips later on `FILE_ATTRIBUTE_NORMAL` fallback and `FILE_ATTRIBUTE_DIRECTORY`-on-file validation — both outside the scope of this fix; left for follow-ups.
- **smb2.async_dosmode** flips from memfs-only-green to passing on cairn + diskfs as a side benefit (the suite reads back attributes after a write and was being thrown by the spurious ctime updates).

Verified end-to-end against `smb2.timestamps`, `smb2.openattr`, `smb2.acls.*`, and the read/rw/share-mode/secleak/notify suites on every native backend — all stay green.

### DIFF
--- a/src/server/smb/smb_attr.h
+++ b/src/server/smb/smb_attr.h
@@ -383,31 +383,48 @@ chimera_smb_unmarshal_basic_info(
     attr->va_req_mask = 0;
     attr->va_set_mask = 0;
 
-    /* Each timestamp is only applied when the client supplied a real value;
-     * the omit/freeze/thaw sentinels leave the stored value alone. */
+    /* MS-FSCC 2.4.7 FileBasicInformation: a non-zero timestamp replaces the
+     * stored value; a zero (or freeze/thaw sentinel) means "don't change".
+     * Carry every time field through to the backend even when it is omitted,
+     * using the TIME_OMIT sentinel — that way the implicit ctime bump (which
+     * apply_attrs would otherwise stamp because the co-present DOS-attribute
+     * change is a metadata write) is suppressed for the fields the client
+     * asked to leave alone.  apply_attrs treats TIME_OMIT as "preserve". */
     if (!chimera_smb_time_is_omit(smb_attrs->smb_atime)) {
         chimera_nt_to_epoch(smb_attrs->smb_atime, &attr->va_atime);
-        attr->va_req_mask |= CHIMERA_VFS_ATTR_ATIME;
-        attr->va_set_mask |= CHIMERA_VFS_ATTR_ATIME;
+    } else {
+        attr->va_atime.tv_sec  = 0;
+        attr->va_atime.tv_nsec = CHIMERA_VFS_TIME_OMIT;
     }
+    attr->va_req_mask |= CHIMERA_VFS_ATTR_ATIME;
+    attr->va_set_mask |= CHIMERA_VFS_ATTR_ATIME;
 
     if (!chimera_smb_time_is_omit(smb_attrs->smb_mtime)) {
         chimera_nt_to_epoch(smb_attrs->smb_mtime, &attr->va_mtime);
-        attr->va_req_mask |= CHIMERA_VFS_ATTR_MTIME;
-        attr->va_set_mask |= CHIMERA_VFS_ATTR_MTIME;
+    } else {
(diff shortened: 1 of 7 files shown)


benjarvis:
Rebased onto upstream/main and amended one fix:

- **The first CI run timed out `smbtorture_smb2_notify_basedir_io_uring` consistently across all 4 devcontainer jobs.** The branch was based on main *before* #620 ("vfs/notify: don't overflow subtree watches on mount-root emits") landed; #620 fixed the underlying notify-overflow on io_uring that this PR was exposing.
- I'd also missed a parallel `utimensat()` call site in `chimera_io_uring_setattr` (the SetAttr path on an already-open fd) — only `chimera_io_uring_set_open_attrs` (the open-time path) had been updated to handle `CHIMERA_VFS_TIME_OMIT`. With `TIME_OMIT` flowing through unfixed, `utimensat` got an out-of-range `tv_nsec` and returned EINVAL, which silently failed the `SetInfo(FileBasicInformation)` from smbtorture's `smb2_util_setatr` — so `notify.basedir` never fired the event it was waiting for. linux uses a shared helper so it already had the fix; io_uring has two separate copies.

`smb2.notify.basedir` now passes on all 6 backends locally; no regressions on `smb2.timestamps`, `smb2.openattr`, `smb2.acls.*`, or any read/rw/notify/share-mode suite I checked.

PR #631, opened 2026-06-02
vfs: compound operations as the mandatory currency of the VFS API (CAP_COMPOUND)

benjarvis:
Reframes the experimental transactions feature as **compound operations** — grouped VFS operations in the style of an NFS4 COMPOUND or SMB2 chain, made the mandatory (never-NULL) currency of the VFS API. `CAP_TRANSACTIONAL` becomes `CAP_COMPOUND`; the semantics move from "optional atomic transaction" to "best-effort grouping token."

## The contract

A compound is a **grouping token, not a transaction**. Its portable obligation is "keep these operations together as well as the owning backend can manage" — nothing stronger.

- **Best-effort, with ejection.** A compound is owned by one `{module, mount}`. An operation that resolves elsewhere (a cross-mount hop, a foreign KV backend) is *ejected* — it runs standalone and its effects are independent of the compound's commit/abort. A proxy that cannot express a pattern on the wire degrades to separate operations the same way. "As well as possible" is understood to not always be 100%.
- **Two lanes, chosen at begin.** *RETRYABLE* keeps the wait-die / conflict-replay machinery (the NFS3 per-RPC driver and the client SDK's per-op driver, whose results never escape before commit). *GROUPING* never delivers a conflict and is the only lane the protocol servers may use — NFS4 seqid/stateid installs, consumed SMB write payloads, and externalized lease breaks make whole-compound replay unsound. Engine conflicts surface to grouping consumers as a retriable per-op error (NFS4ERR_DELAY / SMB `STATUS_FILE_NOT_AVAILABLE` / S3 503 / EAGAIN).
- **Atomicity is permitted, never promised.** An engine *may* apply a compound atomically; a member-op failure never poisons the compound (ending with COMMIT keeps what ran, as NFS4 truncate-at-error and SMB fail-forward require). ABORT is best-effort discard. Visibility/isolation are not promised; the shared caches and change-notify publish per-op, so enlisted effects are provisional until commit.
- **Never-NULL.** Every compoundable VFS op takes a non-NULL compound (real, unbound, or the per-thread LOOSE sentinel for deliberately standalone work); NULL survives only internally as the post-ejection "not enlisted" marker. A debug-only assert enforces this at each proc.

## Backend classes

- **Engines** map a compound to a transaction. **cairn** is the engine of record here: a grouping compound is a RocksDB `WriteBatchWithIndex` (conflict-free, read-your-writes for metadata and file data), a RETRYABLE compound is the optimistic transaction, and a cycle-fold keeps autocommit's amortized-fsync economics (a single client op is one grouped commit, not one sync commit per op). **diskfs keeps its 2PL/wait-die engine dormant** and does not declare the capability — it is deferred to a second wave against the same contract (abort-soundness, per-op lock modes, and no-wait/conflict-on-idle-holder are its entry criteria).
- **Non-compound backends** — memfs, linux, io_uring, and NFSv3 mounts — behave exactly as before: they declare nothing, every op ejects, and the core absorbs begin/end.
- **Proxies** (the NFS4/SMB client modules) are wired to run the grouping profile but declaring `CAP_COMPOUND` per-mount, plus the wire mapping (SMB related-chain fusion, END→COMMIT/FLUSH fan-out), is the follow-on wire phase.

## Consumers

All of them drive compounds: the POSIX layer (one syscall == one compound, with the public caller-compound SDK API and real `OPEN_TRUNCATE` honoring), the client SDK, the NFS3 server (one RPC), the NFS4 and SMB servers (one wire compound/chain), S3 (the replayable metadata phase, with the body streamed standalone), FUSE (one request), and the REST debug endpoint.

## Status

The compound contract is the enforced, observable currency of the VFS API, validated by the quint model-based (MBT) suites across every consumer and by a direct VFS-level contract test; a debug non-atomic cairn mode confirms no consumer relies on the optional atomicity. Landed as reviewable, individually-shippable commits from the rename through the never-NULL flip.

Still to come on this branch: the wire phase (per-mount proxy capabilities, SMB transient-chain fusion, durability fan-out, an optional batch-submission window), and the second wave that brings diskfs back under the contract. Two smaller items are deferred and tracked: notify-gate parking when FUSE sync-watchers coexist with compounds, and NFS4 namespace-mutation durability.

Prior art note: no NAS or distributed filesystem exposes compounds as the mandatory currency of an internal VFS API. API-level filesystem *transactions* have consistently failed (Windows TxF was deprecated and never worked over SMB); best-effort, per-op-status batching (S3, SMB2 chains, NFS4 COMPOUND, io_uring links) is what survived. The rename from "transaction" to "compound" lands on the surviving side of that history.

https://claude.ai/code/session_01XXcZVxo8NFKescfjdy9QAf


benjarvis:
Rebased onto current `main` (`06631ed6`, 201 commits since the previous base). The 18 incremental commits were collapsed into a single clean re-port commit because the conflicts were too intertwined to replay commit-by-commit.

Notable integration work in this rebase:
- Re-homed `CHIMERA_VFS_CAP_TRANSACTIONAL` to bit 25 (main took bit 24 for `CAP_CHANGE`).
- Folded main's secure file-handle decode/wrap + per-export cred squash into the NFS3 transaction driver: each proc decodes the wire FH up front so the transaction routes on the *decoded* handle, and the cross-protocol recall-before-mutate path stashes victim/target handles in `saved_fh`/`rename_target_fh` so a conflict replay can't clobber the routing FH.
- Re-ported the diskfs WFG onto main's rewritten intent-log ack/recycle ring, and fixed a use-after-free + leak the new async apply thread exposed: an explicit (core-owned) txn must outlive `applied_wm` like a pooled autocommit txn, so its EndTransaction completion is deferred from the durable ACK loop to the applied recycle loop (autocommit still acks at durable). Also handled main's new `WRITE_NOPIN` lock mode in the WFG exclusivity checks.

Verification (non-KVM): Release 8197/8199, diskfs ASan 2588/2590 with 0 ASan/leak hits; syntax-check / reuse-lint / copyright-check / scan-build all clean. The only failures are the pre-existing `pjd/chmod/12/nfs3_diskfs_{aio,io_uring}` killpriv-write deadlock (diskfs-only, unrelated to this rebase).

PR #660, opened 2026-06-05
docs: skin Just the Docs to match chimeraproject.io

benjarvis:
## Summary

Retool the Jekyll docs pipeline (`docs/`) so the generated documentation shares the look and chrome of the top-level **chimeraproject.io** site, instead of the stock Just the Docs appearance. This skins Just the Docs rather than replacing it, so we keep search, the auto sidebar nav, TOC, and code highlighting while matching the marquee site.

## What changed

- **New color scheme** `_sass/color_schemes/chimera.scss` — builds on JTD's `dark` scheme (which also brings the github-dark Pygments theme), then re-points the palette at the chimeraproject.io oklch tokens. Surfaces that are only used in plain `color`/`background`/`border` are wired to runtime `var(--token)` (so they honour the `[data-theme]` switch); the handful of vars JTD feeds through Sass colour functions (`darken()`, `rgba()`) stay concrete sRGB conversions of the same oklch values.
- **Typography + tokens** — `IBM Plex Serif` headings, `JetBrains Mono` body/labels, 3px radii (`_sass/custom/setup.scss`), and the full oklch token set incl. a `[data-theme=light]` variant (`_sass/custom/custom.scss`).
- **Shared chrome** — full-width top bar, footer, and the theme/accent/density "Tweaks" panel ported from the main site (`_includes/chimera_topbar.html`, `chimera_footer.html`, `chimera_tweaks.html`, `head_custom.html`), layered into a thin `_layouts/default.html` override kept close to upstream 0.12.0 for easy re-syncing. Cross-site links are absolute to the site root; docs-internal links use `relative_url`.
- **Layout glue** — slot JTD's fixed sidebar/footer underneath the 52px top bar.
- **Config fix** — `_config.yml` had `url:` set to the full `…/chimera` path, which combined with the workflow's `--baseurl /chimera` doubled every `jekyll-seo-tag` canonical/`og:url` (`…/chimera/chimera/`). Set `url:` to the origin and add `baseurl: /chimera` (CI re-passes the same value).

## Verification

- `bundle exec jekyll build` succeeds (only pre-existing JTD Sass deprecation warnings).
- Output contains the tokens/chrome/fonts; github-dark code theme wins the cascade; canonical URLs are correct; `reuse lint` passes (770/770).

## Follow-ups (not in this PR)

- `docs/api.html` (Redoc) is a standalone page still on Redoc's default light theme — can be themed to match.
- The 4 colour-function-bound vars are baked dark-theme hex, so the Tweaks "Light" toggle themes everything except those; can be made `var()`-driven for pixel-clean light mode.

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -5,12 +5,15 @@
 title: chimera
 description: chimera documentation
 theme: just-the-docs
-color_scheme: dark
+# chimera scheme skins Just the Docs to match chimeraproject.io
+# (see _sass/color_schemes/chimera.scss and _sass/custom/).
+color_scheme: chimera
 
+# Origin only; the /chimera path is the baseurl. The Pages workflow re-passes
+# the same value via --baseurl at build time.
+url: https://chimeraproject.io
+baseurl: /chimera
 
-url: https://chimeraproject.io/chimera
-
-aux_links:
-  GitHub: https://github.com/chimera-nas/chimera
-  Discord: https://discord.gg/hnRkvQFecq
+# GitHub / Discord live in the shared top bar (_includes/chimera_topbar.html),
+# so the theme's own aux links are disabled to avoid duplication.
 
--- a/docs/_includes/chimera_footer.html
+++ b/docs/_includes/chimera_footer.html
@@ -1,0 +1,51 @@
+{%- comment -%}
+SPDX-FileCopyrightText: 2026 Chimera-NAS Project Contributors
+SPDX-License-Identifier: Unlicense
+
+Site footer mirroring chimeraproject.io. Marquee pages are absolute (site
+root); the docs API reference is served from this Jekyll build.
+{%- endcomment -%}
+<footer class="site-foot">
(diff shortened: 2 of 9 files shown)


benjarvis:
Follow-up commits on this branch (same docs effort):

- **Add a Configuration page** documenting every JSON config parameter for both the server daemon and the client — `common`, `server` (+ `smb_auth`/`smb_multichannel`/`pnfs`/`vfs` plugin loading), `mounts`/`exports`/`shares`/`buckets`/`users`/`s3_access_keys`, the client `config` section + fio engine options, and per-VFS-module options (memfs, linux, io_uring, cairn, diskfs). Types/defaults verified against `server.c`, `client.c`, `common_config.h`, and each VFS module.
- **Remove the Community page** — the GitHub/Discord links now live in the shared top bar.
- Renumber API Reference so the sidebar reads Quick Start → Building → Configuration → API Reference.

Build green, 19 tables render, REUSE compliant.

PR #667, opened 2026-06-05
smb3: LZNT1 and LZ77+Huffman compression codecs

benjarvis:
Follow-up to #659 (Plain LZ77 + chained Pattern_V1). Adds the remaining two MS-XCA byte codecs so the full SMB3 compression algorithm set is supported.

## What's included
- **LZNT1** (MS-XCA §2.5) — 4 KB chunk framing, flag-byte tagged groups, position-dependent offset/length bit split.
- **LZ77+Huffman** (MS-XCA §2.1) — 64 KB blocks, 256-byte canonical code-length table, byte/bit-interleaved bitstream (2-word-delayed writer).
- **Send-path generalisation** — a READ response is compressed with the client's most-preferred negotiated codec (LZ77 / LZNT1 / LZ77+Huffman); negotiation advertises all three plus Pattern_V1.

Both codecs were written from the public MS-XCA spec and verified bit-for-bit against the reference implementation (decode of MS-produced vectors + our compress → MS decompress).

## Interop subtleties the WPTS suite surfaced (Huffman)
- Symbol 256 is *both* a `(distance 1, length 3)` match and the end-of-stream marker. A trailing such match makes a decoder's "symbol 256 at end of input = EOF" test fire early and drop 3 bytes, so we expand it to literals (as Windows does).
- The 2-word-delayed bit writer's skipped slots are zero-filled (matching the reference), not left uninitialised.

## Validation
Microsoft MS-SMB2 (WPTS) ServerTestSuite: the LZNT1 and LZ77Huffman cases — including the large-file variants — now pass, bringing the compression suite to **39/44**. The 5 remaining skips are the four `*_Encrypted` variants (compression+encryption layering, not yet implemented) and `ChainedCompressionNotSupported` (inapplicable to a chaining-capable SUT). The new cases are added to the `wpts/memfs_compression/*` ctest allowlist.

Unit tests gain LZNT1/Huffman round-trips (incl. multi-block) and decode checks against MS-XCA reference vectors. Full SMB regression (118 tests) green; syntax / reuse-lint / copyright-year clean.


benjarvis:
Superseded by #668, which consolidates this and the parallel codec work into a single best-of. #668 keeps this PR's stronger codec core (the multi-block encoder, the O(1) 15-bit decode table, and the byte-exact MS reference vectors in the unit test), fixes the one scan-build dead-store, and additionally wires the encrypted-codec WPTS variants on top of the compress-then-encrypt layering (#664). Closing in favour of #668 to avoid two competing codec PRs.

PR #690, opened 2026-06-07
s3: ceph/s3-tests compliance suite + bucket lifecycle and crash fixes

benjarvis:
Draft. Wires the [ceph/s3-tests](https://github.com/ceph/s3-tests) S3 compatibility suite into the devcontainer + ctest, adds the minimal server support needed to run it, and fixes the daemon crashes/hangs it exposed. Missing **features** are intentionally not addressed yet.

## Compliance baseline
Per-test-isolated full run of `test_s3.py` (838 cases, memfs): **202 passed / 539 failed / 94 skipped / 3 hung**, and **no daemon crashes**. The core object/bucket data path passes; everything needing an unimplemented feature (versioning, ACL, tagging, lifecycle, SSE, object-lock, policy, CORS, website, post-object, checksums) fails, plus listing gaps and a SigV4 special-character signature bug. The 3 remaining "hangs" are lifecycle tests that `sleep(~70s)` waiting for object expiration chimera doesn't implement (daemon is idle — a feature gap, not a server hang). Full breakdown in `docs/design/s3_ceph_compliance.md`.

## Server changes (enabling)
Runtime bucket operations (previously only static config buckets were served):
- CreateBucket / DeleteBucket / ListBuckets / HeadBucket
- ListObjectVersions — no versioning, so each object is one `null` version (the suite's cleanup uses this)
- recursive DeleteBucket that purges the empty key-path directories left by hierarchical keys (`BucketNotEmpty` only if a real object remains)
- configurable bucket root (`server.s3_bucket_root`, default = first mount)

## Listing (filesystem-mapped)
ListObjects v1/v2 + ListObjectVersions: delimiter/`CommonPrefixes` rollup, lexicographic order, `max-keys`/`IsTruncated`, V1 `marker` + V2 `continuation-token`/`start-after`, `KeyCount`, `encoding-type=url`. Bucket sub-resources we don't implement (acl/policy/lifecycle/encryption/...) are accepted-and-ignored (empty 200) — they have no meaning for a filesystem bucket; security-relevant ones are thus **not enforced** (flagged).

## Robustness bugs fixed (each was a remote-triggerable DoS)
- **Crash:** `GET`/`HEAD` on a key resolving to a directory (or any non-regular file) fatally asserted building an object ETag and aborted the daemon -> now returns `NoSuchKey`.
- **Hang:** an open-ended (`bytes=N-`) or suffix (`bytes=-N`) ranged GET left a `-1` sentinel that wrapped `file_left` to a huge unsigned value, so the read loop spun forever. Now resolved against the object size + clamped to EOF. Fixes boto3 `download_fileobj` of >=8MB objects and the explicit ranged-request tests.
- **`s3_port` config was ignored** (daemon always bound 5000); now honored.
- **ListObjects leading slash** on keys (broke delete-by-key); stripped.

## Harness
- `.devcontainer/Dockerfile` clones the suite into `/opt/s3-tests` (pinned SHA, own venv).
- `src/server/s3/tests/ceph/` — runner, CMake/ctest entry, README, and a 148-test green allowlist.
- ctest `chimera/server/s3/ceph_s3tests` runs the allowlist under a netns as a regression gate (the full suite is a tracking tool, not a gate).

## Not addressed (deliberately)
The ~590 feature-gap failures (versioning, ACL, tagging, lifecycle, SSE, object-lock, policy, CORS, website, post-object, checksums, listing delimiter/pagination, SigV4 special-character keys, `416 InvalidRange`). Follow-up work, tracked in the compliance doc.


benjarvis:
Superseded by decomposed topical PRs: merged #692 (GET robustness), #693 (error XML escaping), #695 (s3_port); and open stack #699 (bucket lifecycle) → #700 (listing) → #701 (ceph/s3-tests harness), all rebased onto current main.

PR #691, opened 2026-06-07
posix: port pjdfstest conformance suite + raise POSIX compliance

benjarvis:
## Summary

Ports the **pjdfstest** POSIX filesystem conformance suite into the Chimera POSIX
client test suite (`src/posix/tests/pjd/<category>/NN.c`, one C program per
upstream `.t`), and raises the POSIX compliance of the client/VFS layers those
tests exercise. Per the design, fixes were made **below the POSIX wrapper** —
in the shared `src/client` and `src/vfs` layers — wherever the behavior is
shared, so NFS and SMB benefit too.

**134 ctest cases** across 14 categories, all green on `memfs`, with no
regressions in the existing POSIX suite (332 memfs cases) or the
`memfs`/`cairn`/`linux`/`nfs` namespace/rename suites.

## Enabling infrastructure

- **Per-call credentials in the client API.** Added `has_cred`/`req_cred` to
  `chimera_client_request` plus a `chimera_client_req_cred()` accessor, routing
  every VFS dispatch through it. The POSIX layer exposes a thread-local override
  (`chimera_posix_set_cred`/`clear_cred`) captured into each request, so a single
  client can act as different users per call — required for the permission tests.
- **umask** support (`chimera_posix_umask`) and a fix for create paths that were
  discarding the requested mode entirely (`open`/`mkdir` did `(void) mode;`).
- New ops: `utimensat`/`futimens` (setattr ATIME/MTIME via the `TIME_NOW`/
  `TIME_OMIT` sentinels) and `pathconf`/`fpathconf`.
- `src/posix/tests/pjd_common.h` — a C analog of pjdfstest's `misc.sh`: TAP
  assertions, a harness working directory, per-call cred/umask switching, file
  creation of every type, name/path generators and stat-field accessors.

## Compliance fixes (shared client/VFS)

- **open** now authorizes the requested access (read for `O_RDONLY`/`O_RDWR`,
  write for `O_WRONLY`/`O_RDWR`, write for `O_TRUNC`); added
  `CHIMERA_VFS_OPEN_WRITE_ONLY` so the access mode is fully represented.
- `O_TRUNC` was unimplemented; opening a directory for writing now returns
  EISDIR; `O_NOFOLLOW` on a symlink returns ELOOP; `O_CREAT` enforces NAME_MAX
  and parent write+search.
- `ENAMETOOLONG` for over-long components/paths (was falling through to ENOENT).
- setattr returns **EPERM vs EACCES** correctly (chmod/chown/explicit-times vs
  size/utimes-now); POSIX **chown** restrictions (non-root can't change uid, gid
  only to a group it belongs to); **setuid/setgid kill-priv** on ownership
  change; **S_ISGID** clearing on non-owner chmod; the owner may always set
  timestamps regardless of write permission.
- Creating/removing an entry requires **search (EXECUTE) + write** on the
  directory; POSIX **delete model** (the NFSv4 per-object DELETE fallback applies
  only to ACL-flavored callers); sticky-bit deletes; `rmdir`/`rename` of
  `.`/`..` → EINVAL/ENOTEMPTY; hard-linking a directory → EPERM; `lchown` no
  longer follows symlinks.
- Timestamp semantics: truncate/chmod/chown/rename bump ctime (and mtime where
  required); unlink of a hard link bumps the surviving inode's ctime.

### Crash/deadlock fixes uncovered by the suite

- `O_NOFOLLOW` create path in memfs left an inode lock held (deadlock).
- `rename` of a directory into its own descendant self-deadlocked; now an
  ancestry walk returns EINVAL before taking the child lock.
- `utimensat`'s directory-relative path used a stack-local attr after scope and
  didn't request the FH (backend abort "no fh returned").
- `ftruncate` on a read-only descriptor now returns EINVAL.

## Scope / deferred

Out of scope for this PR (documented inline): FreeBSD-only rows
(chflags, `st_birthtime`), EROFS/EFAULT/EXDEV/ETXTBSY/ENOSPC/>2GB cases, the
`granular` NFSv4-ACL category, and two single sub-tests (`utimensat/05`
per-component symlink-follow in the dirfd path; `chmod/12` write-clears-SUID).
Tests are registered on `memfs` first; widening to cairn/diskfs/linux/io_uring/
nfs is a follow-up using the existing `generate_posix_backend_tests` machinery.

### DIFF
--- a/src/posix/tests/CMakeLists.txt
+++ b/src/posix/tests/CMakeLists.txt
@@ -947,6 +947,172 @@ foreach(prog test_fcntl test_lockf)
     add_posix_nfs3rdma_test_myuser(${test_name} ${prog} linux)
 endforeach()
 
+# pjdfstest POSIX conformance ports (one program per upstream .t file).
+# Each lives under pjd/<category>/NN.c and shares pjd_common.h.  Registered on
+# memfs first; widen the backend list per category as it greens.
+# ============================================================================
+set(PJD_TESTS
+    open/00
+    open/01
+    open/02
+    open/03
+    open/04
+    open/05
+    open/06
+    open/07
+    open/08
+    open/12
+    open/13
+    open/16
+    open/22
+    open/26
+    chmod/00
+    chmod/01
+    chmod/02
+    chmod/03
+    chmod/04
+    chmod/05
+    chmod/06
+    chmod/07
+    chown/00
+    chown/01
+    chown/02
+    chown/03
+    chown/04
+    chown/05
+    chown/06
+    chown/07
+    truncate/00
+    truncate/01
+    truncate/02
+    truncate/03
+    truncate/04
+    truncate/05
+    truncate/06
+    truncate/07
+    truncate/09
+    truncate/13
+    ftruncate/00
+    ftruncate/13
+    mkdir/00
+    mkdir/01
+    mkdir/02
+    mkdir/03
+    mkdir/04
+    mkdir/05
+    mkdir/06
+    mkdir/07
+    mkdir/10
+    rmdir/00
+    rmdir/01
+    rmdir/02
+    rmdir/03
+    rmdir/04
+    rmdir/05
+    rmdir/06
+    rmdir/07
+    rmdir/08
+    rmdir/11
+    rmdir/12
+    mkfifo/00
+    mkfifo/01
+    mkfifo/02
+    mkfifo/03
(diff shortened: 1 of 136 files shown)


benjarvis:
Superseded by #776 — rebased onto current main and wired the suite in as its own devcontainer CI shard. The head branch now lives on the fork (benjarvis:pjdfstest).

PR #713, opened 2026-06-08
s3: fix response desync on keep-alive connections (double-respond + stale PUT ETag)

benjarvis:
## Problem

S3 responses could desync on reused (keep-alive) connections: a request could observe another request's response or a stale ETag. This is the root cause of the intermittent multipart `InvalidPart` flake and why the ceph `s3-tests` harness had to spawn one daemon per test. Separately, an `Expect: 100-continue` request followed by a reuse of the same connection (`test_100_continue_error_retry`) would hang the client.

## Root cause

Two distinct bugs, both in the chimera S3 server (`src/server/s3/`):

1. **Double response on synchronous VFS backends.** `s3_server_respond()` was reachable twice for a single request. For a PUT, the VFS write/link completion callback (`chimera_s3_put_finish_common`) responds inline, and so does the HTTP `RECEIVE_COMPLETE` notification once the body is in. On a synchronous backend (memfs) the VFS callback runs inline, so **both** paths observe the request ready to send and each appends the full set of response headers and dispatches a status line. The result is a response with **duplicated headers** and a `Content-Length` that no longer frames the body — desyncing every subsequent request on the connection and hanging strict HTTP clients (boto3/urllib3 reports `MissingHeaderBodySeparatorDefect`).

2. **PUT response ETag computed from the pre-write (empty) object.** The PUT ETag was attached in the create / create-unlinked callback, *before* the body was written, so it hashed `(size=0, create-mtime, fh)`. A later HEAD/GET hashes the final `(size=N, write-mtime, fh)` and so returns a **different ETag for every object**, breaking client ETag checks on a reused connection.

## Fix

1. Guard `s3_server_respond()` with a per-request `responded` flag so the status line + headers are emitted exactly once. (GET still streams its body via later `WANT_DATA` notifications, which is why this can't be keyed off `vfs_state`.)
2. Defer the PUT ETag to after the body is written: `getattr` the object's final attributes and attach the ETag from those, matching exactly what HEAD/GET compute.

The interim `100 Continue` handshake was **already** handled correctly by the libevpl HTTP layer (it emits `100 Continue` before reading the body when the client sends `Expect: 100-continue`), so **no libevpl change is required**.

## Before / after (memfs, single long-lived daemon, no per-test restart)

| Check | Before | After |
| --- | --- | --- |
| boto3 keep-alive put/head loop, 200 ops, one pooled socket | 200/200 ETag mismatches | **0/200** |
| ceph/s3-tests curated allowlist (198 cases) on one daemon, x3 | flaky, needed 1 daemon/test | **198/198, stable across 3 runs** |
| `test_100_continue_error_retry` (Expect + connection reuse) | hangs | **passes** |
| chimera `s3_test.py` put/get/head/list/delete/delete_objects/copy/multipart/streaming | pass | **pass** (no regression) |

Debug/ASan build is clean; no daemon crashes under the full load.

> Note on `test_100_continue` (the strict 403-then-100 case): the second leg requires an anonymous-access / bucket-ACL model (allow anonymous PUT after `put_bucket_acl public-read-write`) that chimera's filesystem-backed S3 does not implement, so that specific case is out of scope here. The allowlisted `test_100_continue_error_retry` — which exercises the 100-continue handshake plus connection reuse — passes.

### DIFF
--- a/src/server/s3/s3.c
+++ b/src/server/s3/s3.c
@@ -107,6 +107,19 @@ s3_server_respond(
     char              date_ts[64];
     time_t            now = time(NULL);
 
+    /* The response headers/status are emitted exactly once. Several paths can
+     * race to call this for the same request: for synchronous VFS backends
+     * (e.g. memfs) the VFS completion callback runs inline, so both it and the
+     * HTTP RECEIVE_COMPLETE notification observe the request ready to send.
+     * Responding twice duplicates the response headers and desyncs the
+     * connection, so dispatch the status line + headers only on the first
+     * call. (GET streams its body via later WANT_DATA notifications, which is
+     * why we cannot key this off vfs_state.) */
+    if (request->responded) {
+        return;
+    }
+    request->responded = 1;
+
     strftime(date_ts, sizeof(date_ts), "%a, %d %b %Y %H:%M:%S GMT", gmtime(&now));
 
     evpl_http_request_add_header(request->http_request, "Date", date_ts);
@@ -391,6 +404,7 @@ s3_server_dispatch(
     s3_request->has_part_number    = 0;
     s3_request->op_bucket          = 0;
     s3_request->chunked            = 0;
+    s3_request->responded          = 0;
     s3_request->query_upload_idlen = 0;
     s3_request->query_part_number  = 0;
     s3_request->http_request       = request;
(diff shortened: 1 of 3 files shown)


benjarvis:
Reopened as https://github.com/chimera-nas/chimera/pull/719 from an upstream branch so CI runs (the fork-based head branch didn't trigger the self-hosted runners). Rebased onto main.

PR #745, opened 2026-06-11
diskfs: quiesce in-flight intent-log commits before thread teardown

benjarvis:
Fixes the elbencho `diskfs_verify` CI flake (SIGSEGV at exit in Release; ASan heap-use-after-free in Debug, 2 occurrences).

`diskfs_thread_destroy` drained mtime flushes, background inode drains, and pending block I/O — but nothing covered commits already handed to the intent log whose redo writes were still in flight. The IL's retire path walks each txn's pinned blocks and dereferences `txn->thread` (`diskfs_txn_unpin_blocks` → `diskfs_block_unpin` → `thread->shared`), and posts the CQE into the channel the worker frees at unregister. A teardown racing a late redo write is therefore a use-after-free on both the thread struct and the iq channel. With `unsafe_async` a VFS request completes before its commit is durable, so a caller can legitimately reach thread teardown in exactly that state — which is how elbencho's per-worker client shutdown hit it:

```
READ of size 8 ... in diskfs_block_unpin diskfs.c:2578
  diskfs_txn_unpin_blocks ← diskfs_redo_write_cb ← evpl_block_complete   (IL thread)
freed by diskfs_thread_destroy ← chimera_vfs_thread_destroy ← elb_chimera_worker_destroy
```

Fix:
- **Drain on `commits_inflight` before teardown.** It counts every commit from submission (or SQ-full park) until its CQE is reaped — strictly after the IL's last touch of the txn and the channel — so zero means the channel is quiescent. (The IL consumes an SQ entry and reserves its CQ slot in one step, so `SQ empty && cq_inflight == 0` has no in-between window.)
- The unregister wait pumps the event loop instead of raw-spinning, keeping the CQ draining if anything is ever still in flight there.
- The IL's unregister pass now **aborts loudly** if asked to drop a channel with work in flight, so any future gap in the quiesce protocol becomes a deterministic assert instead of heap corruption.

elbencho diskfs tests looped 20× green under ASan; full posix diskfs battery (664) and smbtorture diskfs suites (110) green.

### DIFF
--- a/src/vfs/diskfs/diskfs.c
+++ b/src/vfs/diskfs/diskfs.c
@@ -9800,9 +9800,21 @@ diskfs_thread_destroy(void *private_data)
      * an elbencho-teardown ASan UAF in diskfs_block_unpin).
      * commits_inflight counts every commit from submission (or SQ-full park)
      * until its CQE is reaped, which is strictly after the IL's last touch
-     * of the txn and the channel, so zero here means quiescent. */
+     * of the txn and the channel, so zero here means quiescent.
+     *
+     * Drain the CQ directly rather than pumping the event loop: the IL posts
+     * commit CQEs WITHOUT ringing the cq_doorbell (it relies on the pinned
+     * worker reaping via the per-iteration poll), and evpl_continue stops
+     * running polls once its poll-iteration budget expires without activity
+     * and parks in the OS wait -- with no doorbell ever coming, that is a
+     * deadlock (seen as mass diskfs test timeouts in CI, where loaded
+     * runners reach teardown with a commit still in flight).  drain_cq also
+     * resubmits commits parked on the SQ-full FIFO, so parked entries make
+     * progress too. */
     while (thread->commits_inflight > 0) {
-        evpl_continue(thread->evpl);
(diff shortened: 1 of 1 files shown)


benjarvis:
The first revision deadlocked diskfs teardown on CI (mass 300s timeouts in the merge-queue batch): the commits_inflight drain pumped `evpl_continue` and relied on the per-iteration cq_poll, but the IL posts commit CQEs **without ringing the cq_doorbell** (it assumes a pinned, actively-spinning worker), and `evpl_continue` stops running polls once its poll-iteration budget expires and parks in the OS wait — no wake source. Locally the final CQE always landed before teardown so the loop was skipped, which is why it never reproduced.

Fixed in 8dc250b7: the wait loop now calls `diskfs_iq_drain_cq` directly (plain CQ-ring read + parked-commit resubmit, no event-loop dependency — progress requires only the IL thread, alive by contract here), and the unregister wait is a sleep-spin again for the same no-doorbell reason. Re-validated: diskfs client tests ×10 loops, elbencho, and the 664-test posix diskfs battery green under ASan.

PR #789, opened 2026-06-13
nfs3: NFSACL sideband protocol (prog 100227) + POSIX.1e ACL translation [draft]

benjarvis:
## Summary

Adds the unofficial **NFSv3 ACL sideband protocol** (NFSACL, RPC program **100227 v3** — Sun/Solaris, supported by the Linux kernel for interop) to chimera, on **both the NFS server and the vfs_nfs client**, bridged to chimera's canonical NFSv4/NT ACL model through a new POSIX.1e translation layer.

The Linux kernel NFSv3 client auto-negotiates NFSACL when the server registers program 100227 (advertised via portmap, multiplexed on the NFS port), so `getfacl`/`setfacl`/`ls` work over NFSv3 against a chimera export.

## What's here

- **Translation layer** — `src/vfs/vfs_acl_posix.{c,h}`: `chimera_acl_to_posix` / `chimera_acl_from_posix`. Reuses the existing deny-aware `chimera_acl_to_mode`/`from_mode` for OWNER@/GROUP@/EVERYONE@ (mode round-trips bit-exact); named USER/GROUP map to ALLOW ACEs; default ACLs become `INHERIT_ONLY|FILE_INHERIT|DIR_INHERIT` templates.
- **XDR** — `src/nfs_common/nfsacl.x` (program 100227 v3), types prefixed `nacl_` to coexist with `nfs3_xdr.h`.
- **Server** — `nfs3_proc_getacl.c` / `nfs3_proc_setacl.c` + registration in `nfs.c`, `nfs_common.h`, `nfs_portmap.c`.
- **Client (vfs_nfs)** — `nfs3_getacl.c` / `nfs3_setacl.c` chained off getattr/setattr, with a per-server capability flag that falls back to mode-synthesis when the upstream doesn't speak NFSACL.

### POSIX.1e fidelity (by design)
POSIX.1e is a **documented-lossy** projection of the canonical NFSv4/NT model: named **DENY** ACEs are dropped, and the POSIX `mask::` is recomputed as the union of group-class perms (chimera stores effective perms; it has no mask concept).

## Verification

- **End-to-end against the real Linux kernel NFSv3 client** (`mount -t nfs vers=3`): `getfacl`, `setfacl -m u:1005:r-x` named-user round-trip (correct `mask::`), and `setfacl -d` default-ACL inheritance into a freshly created child.
- Unit test `vfs_acl_posix_test.c` (all-512-modes round-trip + named/mask/default).
- e2e `src/admin/tests/test_e2e_nfs_acl.py` (root + `acl` pkg; runs in CI).
- Debug + Release clean; `vfs` ctest 11/11; `make syntax` / reuse-lint / copyright / clang-analyze clean.

Three wire-format details were required for the kernel to accept replies: a **double count per ACL** (`xdr_decode_array2` length prefix *plus* an explicit `entries` word — they must match), the **reply mask must cover the request** (empty default ACL still echoes `NFS_DFACL` with a zero-length array), and **no lone `ACL_MASK`**.

## Gaps / parked for follow-up

> Draft — parking this here; the items below are deliberately deferred, not oversights.

1. **Strict POSIX precedence not synthesized.** Named entries are emitted ALLOW-only; no per-named DENY to enforce exact `owner→user→group→mask→other` precedence (a named user can be over-granted by a following GROUP@/EVERYONE@ ALLOW).
2. **Named DENY is a one-way loss** — a named DENY set via NFSv4/SMB is invisible over NFSACL and erased by a later SETACL.
3. **Mask is always recomputed**, so a client-set mask broader than any group-class entry isn't preserved (effective perms are).
4. **Client side is the speculative half.** Works, but adds a round-trip on ACL-bearing getattrs, bounds translated ACLs to `CHIMERA_NFS3_CLIENT_ACL_MAX` (256) before degrading to mode-synthesis, and uses optimistic-then-sticky capability probing with no re-probe.
5. **NFSACL v2 (NFSv2) not implemented** — v3 only.
6. **Count-only query** (`NFS_ACLCNT` without `NFS_ACL`) isn't separately supported; we always return entries when either bit is set (real clients always set both).
7. **`make check` not fully run in-sandbox**: KVM matrix and full `build_clang` (scan-build) pending; `clang --analyze` on the new TUs was clean. Wants a full CI pass before un-drafting.
8. **Default-ACL inheritance corner cases** (NO_PROPAGATE, mode-masking interactions) deserve more coverage.

### DIFF
--- a/src/admin/tests/test_e2e_nfs_acl.py
+++ b/src/admin/tests/test_e2e_nfs_acl.py
@@ -1,0 +1,193 @@
+# SPDX-FileCopyrightText: 2025-2026 Chimera-NAS Project Contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""End-to-end NFSv3 POSIX ACL round-trip test (the NFSACL sideband protocol).
+
+Drives the full pipeline against a live daemon: mount a memfs backend, expose it
+as an NFS export, kernel-mount it over NFSv3, and verify that POSIX.1e ACLs set
+and read with setfacl/getfacl round-trip through the unofficial NFSACL sideband
+protocol (RPC program 100227).  The Linux kernel NFSv3 client auto-negotiates
+NFSACL when the server registers program 100227, which chimera advertises via
+portmap and multiplexes on the NFS port.
+
+memfs stores the canonical (NFSv4/NT) ACL natively, so a named-user entry must
+survive the canonical<->POSIX.1e translation in both directions.
+
+Management operations go through the Python SDK client; the data path uses the
+system mount/setfacl/getfacl binaries.
+"""
+
+import os
+import shutil
+import subprocess
+import tempfile
+
+import pytest
+from chimera_admin import ChimeraAdminError
+
+MOUNT = "aclmount"
+# Export name doubles as the NFS pseudo-path clients mount.
+EXPORT = "/acl1"
+EXPORT_VFS_PATH = "/aclmount"
+
+# Needs root to mount NFS, plus the NFS client and acl tooling.  Skip cleanly
+# elsewhere rather than fail.
+pytestmark = [
+    pytest.mark.skipif(
+        os.geteuid() != 0, reason="kernel NFS mount needs root"
+    ),
+    pytest.mark.skipif(
+        shutil.which("mount") is None, reason="mount(8) not installed"
+    ),
+    pytest.mark.skipif(
+        shutil.which("setfacl") is None or shutil.which("getfacl") is None,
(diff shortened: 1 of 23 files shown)


benjarvis:
I've been thinking about what to do with this and I think I've decided to just shelve it for now, probably forever.   Given that multi-protocol inter-operability is sort of a core principle of the project, trying to do anything with POSIX ACLs seems like the wrong move.   They don't map to Windows or NFS4 ACLs in a nice way, and their real world usage is pretty scant.

PR #866, opened 2026-06-17
smb2: reject lease-break-ack with unknown LeaseKey / wrong client (dirlease negatives)

benjarvis:
## Summary

Rejects a malformed/unauthorized SMB3 directory **lease-break acknowledgment** instead of silently accepting it, greening the two `DirectoryLeasing_Negative_*InLeaseBreakAck` conformance cases.

## Root cause

In `chimera_smb_oplock_break`'s lease path, when `resolve_by_lease_key` found **no** open holding the acknowledged `LeaseKey` on the acking session's tree, the handler fell through to `STATUS_SUCCESS`. MS-SMB2 3.3.5.22.2 requires `STATUS_OBJECT_NAME_NOT_FOUND`.

That single NULL-resolve covers both negative cases:
- **Invalid LeaseKey** — a forged/unknown key matches no open.
- **Invalid ClientGuid** — a *different* client (distinct ClientGuid) connects, opens its **own** tree, and acks with the correct key; that tree doesn't contain the lease holder, so the lookup misses.

## Fix

One guard: in the `is_lease` branch, complete with `STATUS_OBJECT_NAME_NOT_FOUND` when `resolve_by_lease_key` returns NULL, rather than continuing to the success path. (The rest of the block is unchanged — git just re-indented it after removing the `if (open_file)` wrapper.)

## Tests enabled

`DirectoryLeasing_Negative_InvalidLeaseKeyInLeaseBreakAck` and `Negative_InvalidClientGuidInLeaseBreakAck`: xfail → green.

## Verification

- Both targets pass; the `InvalidLeaseState` sibling (which we already rejected correctly) and `BVT_DirectoryLeasing_LeaseBreakOnMultiClients` still pass.
- No regression to file leases: base `BVT_Leasing_FileLeasingV1/V2`, `*_SameLeaseKey`, `BVT_OpLockBreak` all pass (5/5).
- Full dirlease green gate batch: **16/0**; the only remaining dirlease xfail is the deeper `BreakHandleCachingByParentRenamed` (unchanged).
- `make syntax`, reuse-lint, copyright-check clean; Release + Debug build clean.


benjarvis:
Closing: this is a net regression. The change makes a lease-break-ack with no matching open return STATUS_OBJECT_NAME_NOT_FOUND (which the two WPTS dirlease negatives want), but smbtorture smb2.lease.v2_complex2 fails on all 6 backends because it legitimately acks a lease whose open was CLOSED while the break was still outstanding -- Samba/Windows keep that lease in a per-(client,lease-key) table until the ack and return STATUS_OK, whereas chimera releases the lease at close, so the resolve misses and we now wrongly reject it. The two cases ('forged/unknown key' vs 'real lease, open closed') are indistinguishable with chimera's per-tree open lookup. Properly greening the dirlease negatives needs a lease-key table that outlives the open (tracked as future work); reverting here. The dirlease Negative_Invalid{LeaseKey,ClientGuid}InLeaseBreakAck cases stay xfail.

PR #870, opened 2026-06-17
nfs4: delegation-recall correctness (own-I/O self-recall + CLAIM_DELEG_CUR_FH)

benjarvis:
Two related NFSv4 delegation-recall correctness fixes, found via the
`nfstest_delegation` cross-client suite. Together they take the suite from
**1303/1381 to 1775/1777** on this base, with no regressions.

## 1. Don't recall a client's own delegation on its delegation-authorized I/O

A `READ`/`WRITE` authorized by a delegation stateid carries no open handle, so
the server opens the current filehandle on the fly and issues the I/O with no
lease owner. The VFS I/O path then acquires an *implicit* lease under a
synthetic `INTERNAL` owner (`client_key = 0`), whose share probe collides with
the client's own `CACHING` delegation lease and recalls the delegation **the
client itself holds**.

The result is a self-recall: the holder's own `READ`/`WRITE` tears down its own
delegation. Downstream, `nfstest_delegation` sees the delegation recalled
before any conflict, the client falling back to sending `LOCK`s it should
handle locally, and — because the delegation is already gone — the genuine
cross-client `CB_RECALL` no longer firing at the conflicting op.

Fix: carry the holder's own lease owner (NFSv4, the client's `client_id`, the
file-handle hash) on the on-the-fly I/O via `chimera_vfs_{read,write}_owned`,
exactly as the open-stateid path already does. Scoped to the delegation path:
anonymous-stateid and pNFS-DS I/O keep the implicit lease so they still recall
*other* clients' leases.

## 2. Support OPEN with CLAIM_DELEG_CUR_FH

On `CB_RECALL`, a client converts opens it held locally under the delegation
into concrete server open state before `DELEGRETURN`. An NFSv4.1 client does
this with `OPEN(CLAIM_DELEG_CUR_FH)` (the FH-based analogue of
`CLAIM_DELEGATE_CUR`). chimera's claim switch fell through to
`NFS4ERR_NOTSUPP` for it, stalling the recall and preventing a clean
`DELEGRETURN`.

Fix: route `CLAIM_DELEG_CUR_FH` through the existing open-by-FH path (the
delegation-grant gate already declines to grant a *new* delegation for this
claim).

## Validation

`nfstest_delegation` (memfs, NFSv4.1):

| | tests | passed | failed |
|---|---|---|---|
| upstream/main | 1381 | 1303 | 78 |
| + fix 1 | 1695 | 1666 | 29 |
| + fix 2 | 1777 | 1775 | **2** |

The 2 remaining are a narrow cross-client sequencing case ("OPEN (WRITE) reply
to the second client after the READ delegation has been returned"), tracked
separately. No regressions: `nfstest_posix` 461/461, `vfs_state` unit test
80/80, `smb2.oplock.brl{1,2,3}` 3/3.

### DIFF
--- a/src/server/nfs/nfs4_proc_open.c
+++ b/src/server/nfs/nfs4_proc_open.c
@@ -1108,6 +1108,17 @@ chimera_nfs4_open_parent_complete(
             break;
         case CLAIM_PREVIOUS:
         case CLAIM_FH:
+        case CLAIM_DELEG_CUR_FH:
+            /* CLAIM_DELEG_CUR_FH (RFC 8881 §18.16): the client is converting an
+             * open it held under a delegation into concrete open state on the
+             * server, identifying the file by the current filehandle (the
+             * minorversion-1 analogue of CLAIM_DELEGATE_CUR, which carries a
+             * name).  This is exactly an open-by-FH; the delegation-grant gate
+             * already declines to hand out a *new* delegation for this claim, so
+             * the client receives an ordinary open stateid it can then return
+             * the delegation against.  A client issues this in response to a
+             * CB_RECALL, so failing it (NFS4ERR_NOTSUPP) stalls the recall and
+             * prevents a clean DELEGRETURN. */
             chimera_vfs_open_fh(req->thread->vfs_thread, &req->cred,
                                 req->fh,
                                 req->fhlen,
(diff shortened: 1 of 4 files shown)


benjarvis:
Clean before/after on this branch's base (`nfstest_delegation`, memfs, NFSv4.1, KVM harness):

| | tests | passed | failed |
|---|---|---|---|
| upstream/main (baseline) | 1381 | 1303 | 78 |
| this PR | 1695 | 1666 | **29** |

**78 → 29 failures**, +363 net passing assertions. (The test count rises because tests that previously aborted mid-flow — once the holder's own I/O tore down its delegation — now run to completion.)

The remaining 29 are a separate `claim_cur`/second-process recall-response cluster (e.g. "OPEN with CLAIM_DELEGATE_CUR before returning the delegation"), being investigated as a follow-up.

PR #877, opened 2026-06-17
smb2/streams: green smb2.streams.dir and attributes1

benjarvis:
## Summary

Two memfs named-stream (Alternate Data Stream) conformance fixes, each greening a previously-failing `smb2.streams` smbtorture subtest. memfs streams go from 9/14 to 11/14.

### `smb2.streams.dir`
An explicit default data fork named on a **directory** (`DNAME::$DATA`) must be rejected — `NOT_A_DIRECTORY` when the open also carries `FILE_DIRECTORY_FILE`, `FILE_IS_A_DIRECTORY` otherwise. Previously the stream parser collapsed `::$DATA` to "no stream", so the directory was opened as a plain file and the share path returned `SHARING_VIOLATION`.

- `chimera_smb_parse_stream_name` gains an `r_explicit_data_fork` out-param, set in the `stream_len==0 && colon2` branch.
- `chimera_smb_create_open_at_callback` rejects `explicit_data_fork && S_ISDIR` once the open reveals a directory target — this runs before the share-reservation step, so no reorder is needed.
- `memfs_list_streams` no longer reports a default `::$DATA` entry for a directory (a directory has no data fork); the spurious entry made the test client mishandle the `FILE_STREAM_INFO` reply.

### `smb2.streams.attributes1`
The attribute-bearing FILE info classes (`FileBasicInformation`, `FileNetworkOpenInformation`, `FileAttributeTagInformation`, `FileCompressionInformation`, `FileAllInformation`) now require the handle to hold `FILE_READ_ATTRIBUTES` (MS-FSA 2.1.5.11) — a handle opened with only `FILE_READ_DATA` gets `STATUS_ACCESS_DENIED`. Classes carrying no file attributes (standard sizes, internal id, EA size, position, mode, alignment, access, name) are intentionally not gated.

## Verification
- `smb2.streams.dir` and `smb2.streams.attributes1` confirmed green 3× on memfs (Debug + Release).
- No regressions: the existing enabled `smb2.streams.*`, `smb2.getinfo.*`, and `smb2.create.*` suites remain green.
- `make build_release`, scan-build (`build_clang`), `make syntax`, and the copyright-year check all pass.

## Not included (each needs a distinct, larger change — follow-ups)
- `attributes2` — base/stream metadata coherence in the per-fh VFS attr cache.
- `names3` — case-insensitive *path* lookup across memfs (not just stream names).
- `delete` — file-level delete-pending semantics shared across the base file and its streams.

### DIFF
--- a/src/server/smb/smb_internal.h
+++ b/src/server/smb/smb_internal.h
@@ -553,6 +553,11 @@ struct chimera_smb_request {
              * trimmed to the base file.  base_oh is the base file's open handle
              * held across the chained chimera_vfs_open_stream. */
             uint8_t                            has_stream;
+            /* Set when the final component is the explicit default data fork
+             * "file::$DATA" (has_stream stays 0; the base file is opened).  A
+             * directory has no data fork, so the open path rejects this form on
+             * a directory target (smb2.streams.dir). */
+            uint8_t                            explicit_data_fork;
             uint16_t                           stream_name_len;
             char                               stream_name[SMB_FILENAME_MAX];
             struct chimera_vfs_open_handle    *base_oh;
--- a/src/server/smb/smb_proc_create.c
+++ b/src/server/smb/smb_proc_create.c
@@ -1894,6 +1894,22 @@ chimera_smb_create_open_at_callback(
         return;
     }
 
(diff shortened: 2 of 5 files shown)


benjarvis:

PR #881, opened 2026-06-17
smb2/streams: green smb2.streams.attributes2 (stream/base attr coherence)

benjarvis:
## Summary

Greens `smb2.streams.attributes2`, bringing memfs named-stream (ADS) conformance to 12/14. Stacked on #877 (dir + attributes1) — the dir/attributes1 commits dedupe once #877 merges; this PR's net delta is the attributes2 fix.

A named stream shares its base file's metadata (mode, owner, timestamps, DOS attributes) but has its own file handle. Two gaps kept the test red:

1. **Create-time attributes dropped.** The stream-open path passed `set_attr = NULL`, so `FILE_ATTRIBUTE_HIDDEN`/`ARCHIVE` requested at create time were never stamped on the base inode. The create's `set_attr` is now threaded through `chimera_vfs_open_stream`; `memfs_open_stream` applies it to the base inode when the open creates or overwrites the stream.

2. **Stale per-fh attr cache.** The stream fh is distinct from the base fh, so a `SetInfo` on the base (or a sibling handle) refreshed only the base-fh cache entry, leaving the stream fh's cached DOS attributes/timestamps stale. Stream handles are now tagged `CHIMERA_VFS_OPEN_HANDLE_STREAM`, and `chimera_vfs_getattr` bypasses the attr cache for them (no lookup, no refresh) so a stream query always reflects the live base inode metadata.

## Verification
- `smb2.streams.attributes2` green 3× on memfs.
- No regressions: the `smb2.streams.*`, `smb2.getinfo.*`, `smb2.create.*` suites and the pynfs memfs suite (13/13) stay green. The attr-cache bypass is a no-op for non-stream handles.

### DIFF
--- a/src/server/smb/smb_internal.h
+++ b/src/server/smb/smb_internal.h
@@ -553,6 +553,11 @@ struct chimera_smb_request {
              * trimmed to the base file.  base_oh is the base file's open handle
              * held across the chained chimera_vfs_open_stream. */
             uint8_t                            has_stream;
+            /* Set when the final component is the explicit default data fork
+             * "file::$DATA" (has_stream stays 0; the base file is opened).  A
+             * directory has no data fork, so the open path rejects this form on
+             * a directory target (smb2.streams.dir). */
+            uint8_t                            explicit_data_fork;
             uint16_t                           stream_name_len;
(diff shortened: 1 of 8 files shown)


benjarvis:

PR #882, opened 2026-06-17
smb2/streams: green smb2.streams.names3 (case-insensitive opens on memfs)

benjarvis:
## Summary

Greens `smb2.streams.names3`, bringing memfs named-stream (ADS) conformance to 13/14. Stacked on #877 (dir + attributes1) and #881 (attributes2) — those commits dedupe once they merge; this PR's net delta is the names3 fix.

Windows opens are case-insensitive even on a volume that advertises `FILE_CASE_SENSITIVE_SEARCH`. memfs resolved names by a case-sensitive hash, so opening a file via an upper/lower-cased path returned `OBJECT_PATH_NOT_FOUND`.

A case-insensitive directory scan (`memfs_dirent_find_ci`) is added as a fallback when the exact hash lookup misses, in both `memfs_lookup_at` (path components) and `memfs_open_at` (the leaf). It is gated to SMB/Windows callers (`CHIMERA_VFS_AUTH_ATTR`): NFS/POSIX (`AUTH_UNIX`) callers keep strict case-sensitive semantics and never run it. The scan is O(n) in the directory size but is only reached on a miss.

## Verification
- `smb2.streams.names3` green 3× on memfs (Debug + Release).
- No regressions: the `smb2.streams.*`, `smb2.create.*` (incl. `mkdir-dup`, `multi`, `leading-slash`) and `smb2.rename` suites stay green, and the pynfs memfs suite (41/41) is unaffected — case-sensitivity is preserved for NFS by the `AUTH_ATTR` gate.
- `make build_release`, scan-build, `make syntax`, copyright-year all pass.

### DIFF
--- a/src/server/smb/smb_internal.h
+++ b/src/server/smb/smb_internal.h
@@ -553,6 +553,11 @@ struct chimera_smb_request {
              * trimmed to the base file.  base_oh is the base file's open handle
              * held across the chained chimera_vfs_open_stream. */
             uint8_t                            has_stream;
+            /* Set when the final component is the explicit default data fork
+             * "file::$DATA" (has_stream stays 0; the base file is opened).  A
+             * directory has no data fork, so the open path rejects this form on
+             * a directory target (smb2.streams.dir). */
+            uint8_t                            explicit_data_fork;
             uint16_t                           stream_name_len;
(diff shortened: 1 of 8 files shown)


benjarvis:

Issue #1309, opened 2026-06-22
Per-command request StructureSize is not validated for READ/WRITE/CHANGE_NOTIFY/CANCEL/OPLOCK_BREAK (wrong-but-large-enough size accepted instead of STATUS_INVALID_PARAMETER)


benjarvis:
**Severity:** low (high confidence) · **Effort:** ~2 SP (effort:small) · **Audit ref:** V2-267 (2026-06 server-side conformance audit, v2)

## Problem
The codebase establishes a clear pattern of per-command StructureSize validation in 13 of the parsers, but five command parsers omit it and rely solely on the truncation check, which catches a too-short body but not a wrong-yet-sufficiently-long StructureSize.

## Spec
MS-SMB2 3.3.5.2.2 and each command's Receiving section (3.3.5.12 READ, 3.3.5.13 WRITE, 3.3.5.19 CHANGE_NOTIFY, 3.3.5.16 OPLOCK_BREAK ack, 3.3.5.17 CANCEL): 'If the ... StructureSize field ... is not equal to ... the server MUST fail the request with STATUS_INVALID_PARAMETER.' Fixed values: READ=49 (2.2.19), WRITE=49 (2.2.21), CHANGE_NOTIFY=32 (2.2.35), OPLOCK_BREAK ack=24 (2.2.24), CANCEL=4 (2.2.30).

## Code location
`src/server/smb/smb_proc_read.c:259-288; src/server/smb/smb_proc_write.c (chimera_smb_parse_write)`

## Evidence
smb.c:1316 validates only the 64-byte SMB2 header struct_size. Per-command parsers then read request_struct_size (smb.c:1323) and most validate it against the command's fixed value (close.c:669, create.c:5018, negotiate.c:1017, query_directory.c:432, query_info.c:682, and lock/flush/ioctl/echo/set_info/tree_connect/tree_disconnect/logoff/session_setup all have `request_struct_size != SMB2_*_REQUEST_SIZE` checks). A grep across the parsers shows READ, WRITE, CHANGE_NOTIFY, CANCEL, and OPLOCK_BREAK have zero such check; they only reject when the fixed body is truncated via the try_get* `prc` aggregation (e.g. read.c:286-288). A header declaring the wrong StructureSize (e.g. 48 instead of 49 for READ) but supplying enough bytes is accepted and processed.

## Impact
[...]

Issue #1287, opened 2026-06-22
Legacy OPLOCK_BREAK ack accepts a break acknowledgment with no outstanding break (missing BREAKING-state guard present only in the lease path)


benjarvis:
**Severity:** low (medium confidence) · **Effort:** ~2 SP (effort:small) · **Audit ref:** V2-245 (2026-06 server-side conformance audit, v2)

## Problem
chimera_vfs_lease_ack is a no-op on a non-BREAKING lease (line 2413 gate), so no state corruption results, but the legacy path returns STATUS_SUCCESS for an ack that arrives with no break in progress instead of the protocol error the lease path correctly returns. The asymmetry between the two ack paths is the defect.

## Spec
MS-SMB2 3.3.5.22.1: an Oplock Break Acknowledgment is valid only while a break is outstanding on that Open; otherwise the server fails with STATUS_INVALID_OPLOCK_PROTOCOL / STATUS_UNSUCCESSFUL.

## Code location
`src/server/smb/smb_proc_oplock_break.c:637-665`

## Evidence
The lease path guards against a stale/duplicate ack: `if (lease->break_state != CHIMERA_VFS_BREAK_BREAKING) { ... SMB2_STATUS_UNSUCCESSFUL; return; }` (lines 599-603). The legacy path has NO equivalent check on `open_file->grant->lease.break_state`; it only checks `open_file->oplock_break_ack_required` (line 638). After a successful ack, oplock_break_ack_required is cleared (line 665), but a second/spurious legacy ack while the lease is IDLE (e.g. the break already settled via the unacked R→NONE settle path, or via the deadline revoke) still calls chimera_vfs_lease_ack on a non-BREAKING lease.

## Impact
Low: no state corruption, but a non-conforming status code (SUCCESS vs INVALID_OPLOCK_PROTOCOL/UNSUCCESSFUL) for a spurious legacy ack; observable by negative conformance tests.

## Dedup
Closest open issue: none. Distinct from the lease-ack duplicate handling (which IS implemented at line 599) and from the unknown-handle finding above; this is specifically the legacy path lacking the break-outstanding guard its lease sibling has.

---
<sub>Filed from the 2026-06 server-side NFS/SMB/S3 conformance audit (v2, net-new pass). Deduped against all open issues at audit time and independently verified (adversarial verifier; criticals/highs double-verified).</sub>

benjarvis:
Resolved by PR #1321 (merged).

Issue #1286, opened 2026-06-22
Lease break ACK resolves the lease key only within the connection's own tree, not the per-client lease table


benjarvis:
**Severity:** low (low confidence) · **Effort:** ~3 SP (effort:medium) · **Audit ref:** V2-244 (2026-06 server-side conformance audit, v2)

## Problem
Because chimera scopes the lookup to request->tree, a lease ack must arrive on the exact tree the lease was created on. Windows clients normally do ack on the same connection, so this rarely bites, but a multichannel client or one that opened the file via a different share to the same volume could ack on another tree and the server would silently treat it as an unknown lease (returning SUCCESS while leaving the lease BREAKING until the deadline).

## Spec
MS-SMB2 3.3.5.22.2 + 3.3.1.x (Global lease table): a lease is identified by LeaseKey within the client's lease table (ClientGuid scope), independent of which tree connect or session channel the acknowledgment arrives on. The server must locate the lease in the per-client LeaseTable, not a single tree.

## Code location
`src/server/smb/smb_internal.h:2641-2668`

## Evidence
chimera_smb_open_file_resolve_by_lease_key iterates only `tree = request->tree`'s open_files buckets (line 2646, 2651-2664). A lease key is a per-client construct (the LeaseTable is keyed by ClientGuid+LeaseKey per MS-SMB2 3.3.1.x), but the resolver scope is a single TreeConnect. The break notification itself is delivered on `open_file->create_conn` (the creating connection), so an ack arriving on a different TreeConnect of the same session (or a different channel/connection of the same client) will not find the open and falls into the not-found→SUCCESS path above.

## Impact
Low/narrow: lease break may be left unacknowledged (and only torn down by the deadline-revoke / next conflict) if the client acks on a different tree connect than the create. Interacts with #1030 to extend the BREAKING window.

## Dedup
Closest open issue: #1030 (break left BREAKING). #1030 is the general no-proactive-timer issue. This is a specific resolver-scope defect (tree-local vs client-global lease table) that can cause valid acks to be missed entirely; a correct global lease lookup would resolve and settle them.

---
<sub>Filed from the 2026-06 server-side NFS/SMB/S3 conformance audit (v2, net-new pass). Deduped against all open issues at audit time and independently verified (adversarial verifier; criticals/highs double-verified).</sub>

benjarvis:
Resolved by PR #1321 (merged).

Issue #1285, opened 2026-06-22
Legacy oplock break ACK re-grants a read cache (LEVEL_II) the server never offered — no subset check against break_needed_mode


benjarvis:
**Severity:** low (medium confidence) · **Effort:** ~2 SP (effort:small) · **Audit ref:** V2-243 (2026-06 server-side conformance audit, v2)

## Problem
When a legacy EXCLUSIVE/BATCH oplock is broken by a write or namespace invalidation, chimera_smb_lease_break_cb computes new_level=NONE (new_vfs has no R bit) and sets oplock_break_ack_required=true (smb_proc_oplock_break.c:349-358). break_needed_mode is set to 0 (one-shot floor). A client (buggy or hostile) can then send an OPLOCK_BREAK ack with OplockLevel=SMB2_OPLOCK_LEVEL_II. The legacy ack handler maps II→CHIMERA_VFS_LEASE_MODE_R and lease_ack stores granted=R, re-arming a READ cache on data that a conflicting writer is mutating — a cache-coherence violation. The lease path explicitly guards against exactly this; the legacy path does not.

## Spec
MS-SMB2 3.3.5.22.1 (Processing an Oplock Acknowledgment, server-side): the server must resolve the outstanding break to a level no higher than the break target; an OplockLevel in the ack that exceeds the level the server broke to is invalid. A break driven by a write/namespace invalidation breaks to NONE (break_needed_mode=0).

## Code location
`src/server/smb/smb_proc_oplock_break.c:653-660`

## Evidence
In the legacy (non-lease) ack path, the kept VFS mode is derived purely from the client-supplied OplockLevel: `kept.granted = (oplock_break.oplock_level == SMB2_OPLOCK_LEVEL_II) ? CHIMERA_VFS_LEASE_MODE_R : 0;` and then `chimera_vfs_lease_ack(&open_file->grant->lease, kept)` is called. chimera_vfs_lease_ack (src/vfs/vfs_state.c:2414 `lease->mode = resulting;`) stores the kept mode VERBATIM with no clamp. Unlike the lease-variant path (smb_proc_oplock_break.c:610 `if (kept_vfs & ~lease->break_needed_mode) { ... SMB2_STATUS_REQUEST_NOT_ACCEPTED }`), the legacy path performs NO validation that the acked level is a subset of what the break asked the holder to retain (lease->break_needed_mode).

## Impact
[...]

PR #1321, opened 2026-06-23
smb2/vfs: oplock & lease-break conformance overhaul (+ NFS3 cross-protocol recall)

benjarvis:
## Summary

Overhauls the SMB2/VFS oplock & lease-break path for protocol conformance, plus the cross-protocol break-before-mutate invariant on the NFSv3 side. Closes a cluster of 2026-06 audit issues and greens a set of previously-disabled smbtorture cases and one WPTS case.

### Break-ack handler correctness
- Reserved `LeaseFlags` in the break-ack reply forced to 0 (#1282)
- Lease-ack resolved across the whole session's trees, not just `request->tree` (#1283, #1286)
- Legacy ack brought to parity with the lease path: BREAKING-state guard (#1287); reject an acked level above the break target or a re-granted LEVEL_II never offered (#1284, #1285)
- A 52-byte v2 RqLs is honored only on SMB 3.x; 2.0.2/2.1 get a v1 lease (#1281)
- Per-command `StructureSize` validation for READ/WRITE/CHANGE_NOTIFY/CANCEL (#1309)

### Break lifecycle / proactive revoke
- The idle reaper now force-revokes a lease stuck `BREAKING` past its deadline even with no new contender — a dead-but-not-disconnected client can no longer pin caching state (#1030)
- `FILE_OPEN_REQUIRING_OPLOCK` fails the create with `STATUS_CANNOT_BREAK_OPLOCK` when the oplock/lease cannot be granted (#1016)

### Cross-protocol break-before-mutate (NFSv3)
- NFSv3 REMOVE/RMDIR/RENAME resolve the victim/clobbered FH and recall any held delegation/lease before the mutation (#1070, #1071)

### CREATE-context robustness
- A context chain > 1024 bytes is heap-parsed instead of silently dropped (#1115)
- The present-mask bit is set only after the body validates, so a malformed RqLs/DH2Q can't honor stale pooled fields (#1116)
- Multi-element byte-range lock `cb_private` points at the open's caching grant, restoring the same-owner exemption (#1252)

### Oplock/lease grant + break semantics (re-enables disabled smbtorture cases)
- Attribute-only ("stat") opens acquire their requested oplock when sole and break nobody otherwise; legacy oplocks cap-to-coexist behind a read holder instead of breaking it — `smb2.oplock.batch9`, `batch9a`
- A lease state with W/H but no R is granted NONE (`smb2.lease.request`); named-stream oplock/lease subtests get ADS enabled in the harness (`smb2.oplock.batch26`)
- Delete-on-close via SET_INFO marks the file delete-pending → later opens get `STATUS_DELETE_PENDING` (`smb2.oplock.doc`)
- A share-conflicting open handle-breaks a conflicting RqLs lease RWH→RW (`smb2.lease.break_twice`)

### WPTS
- Promote `ResilientOpenScavengerTimer_ReconnectAfterTimeout` to green (now passes via the break-deadline revoke).

## Verification
All changes regression-gated against the enabled smbtorture oplock/lease/durable/dirlease/create/streams/session suites (memfs + linux), the WPTS base/persistent/dirlease buckets, NFS3 cthon (memfs), and the phase0 unit tests; `make syntax`/`reuse-lint` clean; debug + release builds clean.

### DIFF
--- a/src/server/nfs/nfs3_proc_remove.c
+++ b/src/server/nfs/nfs3_proc_remove.c
@@ -5,9 +5,24 @@
 #include "nfs3_procs.h"
 #include "nfs_common/nfs3_status.h"
 #include "nfs_common/nfs3_attr.h"
+#include "server/server.h"
 #include "vfs/vfs_procs.h"
 #include "vfs/vfs_release.h"
 #include "nfs3_dump.h"
+
+/* True when a cross-protocol caching holder (NFSv4 delegation, SMB lease, or
+ * SMB oplock) could exist on the shared store.  An NFSv3 REMOVE/RMDIR of an
+ * object another protocol holds must recall that holder before the unlink
+ * (RFC 7530 §10.4.5 / MS-SMB2 break-before-mutate, #1070); when no caching
+ * protocol is enabled the extra LOOKUP needed to learn the victim FH is
+ * skipped. */
+static inline int
+chimera_nfs3_recall_needed(struct chimera_server_nfs_thread *thread)
+{
+    return chimera_server_config_get_nfs4_delegations(thread->shared->config) ||
+           chimera_server_config_get_smb_leases(thread->shared->config) ||
+           chimera_server_config_get_smb_oplocks(thread->shared->config);
+} /* chimera_nfs3_recall_needed */
 
 static void
 chimera_nfs3_remove_complete(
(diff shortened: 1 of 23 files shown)


benjarvis:
Rebased onto current main (resolved conflicts in nfs3 rename — secure-FH decode, wpts_cases.csv — upstream already promoted both ScavengerTimer cases, and the create-context ExtA handler). Submodule pointers left untouched (match main).

Added one more verified fix: **smb2.lease.v2_rename** — the rename source recall is now a single handle recall (RH→R) instead of a cascade-to-NONE, so a v2 lease sees one break, not two. NFSv4 delegation recall on rename is unaffected (recalled explicitly by the NFSv4 rename op before rename_at) — verified by the nfstest_delegation KVM suite. Post-rebase: 208/208 enabled oplock/lease/durable/dirlease/create/session smbtorture (memfs) green.

benjarvis:
Rebased onto current main (64 commits of drift; the only conflict was a comment-wording difference on the lease break-ack Reserved-Flags fix — upstream landed the same #1282 fix independently, kept main's wording). Submodules untouched.

Extended with three more verified break-sequence fixes (new commit `82a1b37`):
- **smb2.lease.lock1** — a byte-range lock breaks a conflicting caching lease to NONE in a single shot (was leaving H and cascading into two notifications).
- **smb2.lease.lease-epoch** — a fresh lease requested behind another owner's range lock is capped to NONE, and a v2 lease granted NONE echoes its requested epoch unchanged.
- **smb2.oplock.batch22a** — break-timeout duration assertion passes by advertising chimera's 30s break-revoke deadline (`torture:oplocktimeout=30`).

Range-vs-caching exemptions preserve same-owner/same-grant locks and an NFSv4 client's own lock vs its own delegation. Verified post-rebase: 147/147 enabled oplock/lease/durable/dirlease/create/lock smbtorture (memfs) + nfstest delegation/lock KVM (NFSv4 unaffected).

PR #1350, opened 2026-06-25
diskfs: punch holes on block boundaries so DEALLOCATE never leaks free space

benjarvis:
## Problem

A diskfs `DEALLOCATE` (`FALLOC_FL_PUNCH_HOLE`, and the `DEALLOCATE` half of the
NFSv4.2 `FALLOC_FL_ZERO_RANGE = DEALLOCATE + ALLOCATE` pair that the Ubuntu
26.04 client issues) trimmed an extent at the exact sub-block byte boundaries
the client requested, leaving a **sub-block device extent whose edge 4 KiB block
was shared with the freed region**.

The space map only ever credits whole 4 KiB blocks *strictly contained* in a
freed byte range (`sm_free_block_range` aligns the start up and the end down), so
the sub-block edge of every punch — and any block shared by two sub-block
siblings — was never credited back. Over many sub-block punches `free_bytes`
drained **monotonically** until a write or punch hit a **spurious `ENOSPC` on a
near-empty device** (issue #733).

Measured on the `chimera/vfs/zerorange_diskfs_*` reproducer (1 GiB device, 256
KiB file), on top of the already-merged #1345 (the separate alloc-path
false-ENOSPC race) so this is genuinely the residual leak, not that race:

| | net_used after 20 000 sub-block ops | first spurious ENOSPC |
|---|---|---|
| before | 266 KB → **53.7 MB** (unbounded) | ~63 000 ops |
| after  | stays **~205 KB** (bounded) | none through 400 000 ops |

## Fix

Split a punch so **no 4 KiB block is ever shared between a freed region and a
surviving extent** — the invariant chosen in lieu of a per-block refcount (no new
persisted structure):

- The **block-aligned interior** of the hole is removed and freed *whole*, so the
  space-map accounting can no longer drop an edge. With the hole now block
  aligned, every surviving head/tail extent is block aligned in both file and
  device offset, and the frees round up to whole blocks.
- The two **sub-block remainders** are zeroed in place on their backing device
  block by a read-modify-write (`diskfs_dealloc_edge_step`), leaving them owned by
  the surviving extent and reading as zeros — rather than carved into a sub-block
  extent that shares a block. (`ZERO_RANGE` keeps no space, so this in-place
  absorb suffices; for a large punch the interior is still reclaimed whole.)

The only device source of sub-block extents was this trimming path, so the fix is
contained to it (`diskfs_io.c`); the space-map free accounting is unchanged.

### Crash consistency

The edge zero-writes are *data* and complete before the dealloc transaction's
redo record commits (`op_ok` runs only at the end of the dealloc chain), carrying
FUA in the crash-safe mode (`unsafe_async` drops it exactly as the ordinary write
path does). So a crash replays to a fully-punched or un-punched range, and never
to a punched extent whose edge still reads stale bytes. No new allocation, copy,
or free rides the path beyond the in-place zeroing, so there is no torn
copy-out / double-free to replay.

## Validation

- `net_used` stays bounded (~205 KB) across **400 000** sub-block ops with no
  `ENOSPC` (was draining to ENOSPC by ~63 000).
- `chimera/vfs/zerorange_{memfs,cairn,diskfs_aio,diskfs_io_uring}` pass, including
  a 10 000-op run with per-op read-back and a **sub-block-aligned file size**
  variant (final extent ends mid-block).
- **384** parallel isolated diskfs runs (16-way) clean.
- `clone_range`, `statfs_mask`, `state`, the diskfs space-reclaim posix tests,
  elbencho data-verify, s3 diskfs data tests, and pynfs `combined_diskfs`
  (4.0/4.1/4.2, aio + io_uring) all green under ASan.
- `make syntax-check` clean; clang static analyzer clean on `diskfs_io.c`.

Completes the partial fix in #1328 (the residual free-accounting leak); refs #733.

### DIFF
--- a/src/vfs/diskfs/diskfs_internal.h
+++ b/src/vfs/diskfs/diskfs_internal.h
@@ -264,6 +264,19 @@ struct diskfs_request_private {
     int                         loop_have;
     uint64_t                    alloc_cap;   /* ALLOCATE: adaptive per-chunk cap */
 
+    /* DEALLOCATE edge-zeroing (diskfs_dealloc_edge_*): the punch's two sub-block
+     * remainders -- [hole_start, SM_ALIGN_UP(hole_start)) and
+     * (hole_end & ~MASK, hole_end) -- are zeroed in place on the backing device
+     * block so the surviving extent owns whole blocks (no block is ever split
+     * between a freed region and a kept extent, which is what keeps the space-map
+     * free accounting whole-block and leak-free).  dz_zs is the moving cursor
+     * over the remainders (the block-aligned interior is skipped); dz_ze is the
+     * hole end; dz_edge is the current block's RMW phase (0 = read, 1 = write);
+     * dz_blk is the 4 KiB RMW buffer for the block being zeroed. */
+    uint64_t                    dz_zs, dz_ze;
+    int                         dz_edge;
+    struct evpl_iovec           dz_blk;
+
     struct evpl_iovec           iov[66];
 
     // For RMW (read-modify-write) on partial block writes
--- a/src/vfs/diskfs/diskfs_io.c
+++ b/src/vfs/diskfs/diskfs_io.c
@@ -415,6 +415,14 @@ diskfs_dealloc_process(
     struct chimera_vfs_request *request);
 
 static void
+diskfs_dealloc_edge_step(
+    struct chimera_vfs_request *request);
+
+static void
+diskfs_dealloc_loop_start(
+    struct chimera_vfs_request *request);
+
+static void
 diskfs_dealloc_first_cb(
     struct diskfs_bt_op *op,
     int                  result,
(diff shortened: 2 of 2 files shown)


benjarvis:
Reviewer note — interaction with #1328: this re-enables `SM_ALIGN_UP` on the dealloc free calls in `diskfs_io.c` (the "completely inside" and "overlap-start" cases), which #1328 had changed to raw-length to avoid over-freeing a 4 KiB block shared with a kept sibling extent. That is now safe **because** the edge-zeroing makes every surviving extent block-aligned, so no block is ever shared — the only sub-block extent that can remain is a file's own last extent, which owns its trailing block exclusively. The inline comment at that hunk explains it, but calling it out here since it's the key coupling to verify. The 10 000-op per-op-readback zerorange run + the sub-block-aligned-file-size variant (both green under ASan) exercise exactly the corruption #1328 fixed.

PR #1491, opened 2026-07-10
nfs4: fix open/lock owner use-after-free against the lease sweeper

arenaud16:
Running pynfs over NFSv4.0 against a blocking (CHIMERA_VFS_CAP_BLOCKING) backend crashes under AddressSanitizer: an OPEN resolves its open_owner at dispatch and stashes a borrowed, unrefcounted pointer on the request while the open goes async into the VFS.  If the client's lease lapses during the round-trip, the lease sweeper frees every owner in nfs_client_expire_state, and the OPEN completion writes owner->seqid into freed memory.  Blocking backends dispatch to a sync delegation thread, stretching the async window from same-tick (memfs, inline) to milliseconds, which is what makes the 1 Hz sweeper race observable.

Fix the lifetime and the surrounding install-vs-expiry races:

- Refcount nfs_open_owner and nfs_lock_owner like every sibling state object: find_or_create returns a caller reference taken under client->lock, in-flight OPEN/LOCK requests hold it across the async VFS round-trip, teardown unpublishes the owner and drops only the hash-slot reference, and the last put() frees.  get/put abort on underflow.
- Pin the owner from open_state/lock_state for the state's lifetime, and pin the open_state from the lock_state (nfs_lock_state_destroy dereferences it with nothing else keeping it alive).
- Serialize open_state/lock_state installation against nfs_client_expire_state under client->lock: creating state on an owner the sweep already unpublished (or a destroyed open_state) now fails and the operation returns NFS4ERR_EXPIRED, instead of installing orphaned state that no teardown walk can ever reach (permanently leaking a dead client's byte-range lock) or tripping the leftover-lock-states abort.
- Keep NFSv4.0 seqid/replay state on a single owner object across a sweep: the OPEN completion re-adopts (republishes) the request's pinned owner rather than creating a fresh one, so a retransmit still finds the cached reply instead of re-executing (e.g. a GUARDED create returning a spurious NFS4ERR_EXIST).
- Make the two lock_state list unlinks idempotent with flags checked under the respective locks: LOCKU racing the expire/CLOSE cascade previously double-unlinked the utlist entries and corrupted the lists.  Encode the stateid of a freshly-created state before dropping the locks; after publication a concurrent sweep can already have freed it.

Add test_open_owner_lifetime covering the borrow-across-sweep cases, the install-after-expiry refusals, full teardown of a linked lock_state, owner re-adoption, refcount abort diagnostics, and a threaded worker-vs-sweeper stress run under ASAN that reproduces the original interleavings.

### DIFF
--- a/src/server/nfs/nfs4_proc_lock.c
+++ b/src/server/nfs/nfs4_proc_lock.c
@@ -91,6 +91,18 @@ chimera_nfs4_lock_finish(
         }
     }
 
+    /* Drop the owner borrow refs transferred onto the request during
+     * dispatch.  Unconditional so every status/minorversion path releases
+     * them exactly once; NULL to guard against a double drop. */
+    if (req->lock_4_0_open_owner) {
+        nfs_open_owner_put(req->lock_4_0_open_owner);
+        req->lock_4_0_open_owner = NULL;
+    }
+    if (req->lock_4_0_lock_owner) {
+        nfs_lock_owner_put(req->lock_4_0_lock_owner);
+        req->lock_4_0_lock_owner = NULL;
+    }
+
     chimera_nfs4_compound_complete(req, status);
 } /* chimera_nfs4_lock_finish */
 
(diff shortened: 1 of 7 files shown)


arenaud16:
Oops there is an issue with this commit

PR #1500, opened 2026-08-04
smbtorture: make the matrix a retainable measurement

pdunlap:
## Why

`tools/smbtorture/regen.sh` could do exactly one thing: run the whole 3930-cell
backend × subtest matrix and immediately collapse it into the
`SMBTORTURE_ENABLED_*` allowlists in `src/server/smb/tests/CMakeLists.txt`. The
evidence landed in `/tmp` and `/var/tmp`, where the next run overwrote it.

Because `add_smbtorture_backend_tests()` registers a ctest only for the *enabled*
set, every currently-failing subtest is invisible to ctest. So there was no way
to ask the two questions that matter for working the backlog down:

- what is still failing?
- has that set changed since last time?

This series makes the measurement a first-class output, diffs it against the
checked-in baseline, and puts both behind a nightly job.

## What changed

**`regen.sh` — the measurement is retainable and scopable**

| Flag | Effect |
|---|---|
| `--report-only` | run the matrix, skip the allowlist splice |
| `--out-dir <dir>` | `results.txt` + per-failure logs land here |
| `--backends <list>` | narrow to a subset of backends |
| `--filter <regex>` | narrow to matching subtests |
| `--build-dir <dir>` | locate `smbtorture_test` explicitly |

Splicing from a `--backends`/`--filter` run is refused outright — a partial
matrix would silently drop every unmeasured subtest from the allowlists.

The build tree is now auto-detected across `./build/Release` (worktrees) and
`/build/Release` (main tree). The previous hardcoded relative path meant the
script could not run in the main tree at all.

**`regen.py` — two named modes**

`generate` is the previous behaviour; `compare` buckets a run against the
checked-in baseline. `--output` writes somewhere other than the input
CMakeLists, `--results` may be repeated (a cell whose status disagrees between
runs is reported as flaky), and `--fail-on-newly-failing` exits non-zero only for
a regression in a gated test.

`compare` buckets on the runner's own classification rather than on "not in the
allowlist", which matters in two places: a 300s hang (`TIME`) is never filed as a
plain failure, and a subtest that is in the allowlist but excluded by
`SMBTORTURE_DISABLED_SUBTESTS` is reported separately — it is not a gated test,
so it must not trip the regression gate.

**`SMBTORTURE_EXPLORE_TESTS=ON` — the complement, addressable from ctest**

Registers, per backend, the subtests the gating suite does not run: the
complement of the generated allowlist plus the entries held back by
`SMBTORTURE_DISABLED_SUBTESTS`. The `smbtorture` and `smbtorture_explore` labels
together now cover the catalog exactly once per backend (933 explore entries
across six backends = the 879-cell complement plus 9 disabled subtests each), so

```
ctest -L smbtorture_explore -R smb2_streams
```

iterates on one family without hand-running the matrix.

The disabled subtests are included deliberately: they are excluded from gating
because they flake, which is a reason to keep them addressable, not a reason to
make them unreachable.

Explore entries mirror `_smbtorture_add_subtest`'s properties — `SKIP_RETURN_CODE
77`, 300s timeout, the durable-family `RESOURCE_LOCK`, the LSAN suppressions — so
a subtest promoted into an allowlist behaves identically before and after. Off by
default; with the option off the registered ctest set is unchanged (verified:
8135 tests, name-for-name identical to before the series).

**Nightly job `smbtorture matrix (informational)`**

Runs the full matrix with `--report-only`, publishes `results.txt` and the
per-failure logs as an artifact, and renders the compare buckets into the job
summary.

It fails only on a *newly* failing cell — a gated subtest that stopped passing. A
*still* failing cell is the expected state of the exploration backlog; gating on
it would make the job red every night and nobody would read it. The verdict is
recorded as a flag file inside the container rather than by failing the compare
step, so the summary and artifacts publish even on a regression, which is exactly
when they are worth reading.

**Manual scoping, so the first real run isn't a blind three-hour bet**

The matrix job could initially only be exercised as the full sweep, so the first
run costs hours and a trivial mistake — a bad artifact path, a mis-quoted flag —
costs another. Three `workflow_dispatch` inputs scope it, defaulting to the full
matrix so the scheduled run is unchanged:

```
gh workflow run nightly.yml --ref <branch> \
    -f smbtorture_backends=memfs -f smbtorture_filter='smb2\.streams'
```

These reach `regen.sh` as `SMBTORTURE_BACKENDS` / `SMBTORTURE_FILTER` /
`SMBTORTURE_PARALLEL`, passed through by `docker run -e NAME` from the step
environment rather than interpolated into the container command line. A dispatch
input is attacker-controlled text and must not be able to break out of quoting.
`regen.sh` grows env fallbacks for the two it lacked, which also makes scoping
available to callers that cannot easily append flags.

## Reporting fixes surfaced along the way

Three, all on the scoped path, none reachable from the full-matrix run:

- `compare` printed the whole allowlist over the measured catalog, so a scoped
  run reported nonsense like "baseline 551/15". It now reports `baseline n
  catalog` and measures against the catalog the run actually covered — the
  filtered list when scoped, else the catalog `regen.sh` just discovered from the
  live smbtorture, rather than the checked-in `SMBTORTURE_SUITES`. Identical
  today; the live list is the honest denominator once a Samba upgrade moves the
  catalog.
- An empty bucket set was read as "no cells measured", but buckets only hold
  deviations — so a backend whose every measured cell matched the baseline, the
  expected outcome, reported as though nothing had run. It now reports the
  measured count and says "no drift from the baseline".
- Backends a scoped run skipped entirely each got their own empty section,
  burying the ones actually measured. They now collapse into a single "not run"
  line.

The full-matrix path (no `--subs`) is unchanged at 551/655.

## Note on `generate` and hand-written commentary

`generate` previously destroyed hand-written rationale.
`SMBTORTURE_ENABLED_memfs` carries comments against individual entries, and
`emit_block()` only ever wrote `# smb2.<suite>` group headers — so any regen
deleted all of them. The entries survived; the reasoning did not. Comments
attached to an entry are now carried onto it across the rewrite.

One consequence worth flagging for review: because `generate` re-sorts the list,
prose covering several adjacent entries cannot survive as a group, so the three
entries that shared one comment each get their own copy in this PR. One comment
per entry is the only association a rewrite can preserve, and it makes the parse
a fixed point.

The CMake lists are parsed line by line rather than with a single non-greedy
regex, which mis-pairs its parens with `SMBTORTURE_RPC_SUITES_DISABLED` — that
list closes on its last entry rather than on its own line — and swallows
`SMBTORTURE_SUITES` entirely.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -505,6 +505,111 @@ jobs:
               /workspace/scripts/cap_ctest_output.sh \
               ctest --output-on-failure --test-output-size-failed 65536 --timeout 300 -j 32'
 
+  # Informational measurement of the full smbtorture smb2 matrix (655 subtests x
+  # 6 VFS backends).  The gating suite only registers the subtests in the
+  # generated SMBTORTURE_ENABLED_<backend> allowlists, so 15-30% of the catalog
+  # per backend never runs in CI at all; this job runs the complement too and
+  # diffs the outcome against the checked-in baseline.
+  #
+  # It fails ONLY on a "newly failing" cell -- a subtest that is in the baseline
+  # (and therefore gated) but did not pass.  A "still failing" cell is the
+  # expected state of the exploration backlog and must not be noise, or the job
+  # would be red every night and nobody would read it.
+  #
+  # amd64 Release only: the matrix is ~3900 daemon+netns+client invocations and
+  # the baselines were measured on that configuration.
+  smbtorture-matrix:
+    name: smbtorture matrix (informational)
+    needs: [build-devcontainer]
+    runs-on: arc-amd64
+    permissions:
+      contents: read
+    # Passing cells cost ~0.6s, but a hung cell burns the full 300s timeout and
+    # there are ~880 non-passing cells across the matrix; 3h leaves headroom.
+    timeout-minutes: 180
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v5
+        with:
+          submodules: recursive
+
+      # --privileged: the netns wrapper needs CAP_NET_ADMIN.  smbtorture itself
+      # ships in the devcontainer image (samba-testsuite).
+      # The matrix and both compare passes run in the container (which is where
+      # python3 and smbtorture live); the runner steps only read the files it
+      # leaves behind, so publishing never depends on tooling outside the image.
+      #
+      # compare runs twice over the same results: once to render the buckets for
+      # the job summary, once to decide the exit status.  The verdict is recorded
+      # as a flag file rather than failing here, so the summary and the artifacts
+      # still publish when a gated test regressed -- exactly when they are worth
+      # reading.  "still failing" is summarized but not listed; it is ~880 cells
+      # of known backlog.
+      - name: Build and run the matrix
+        run: |
+          rm -rf smbt-report
+          mkdir -p smbt-report
+          docker run --rm --privileged \
+            -e DOCKER_MIRROR=${{ env.DOCKER_MIRROR }} \
+            -e APT_MIRROR=${{ env.APT_MIRROR }} \
+            -v "${{ github.workspace }}:/workspace" \
+            -v /build \
+            -w /build \
+            ${{ env.DEV_IMAGE_TAG_PREFIX }}:${{ github.sha }}-amd64 \
+            bash -euxc '\
+              cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -DREQUIRE_NETNS_TESTS=ON /workspace && \
+              ninja smbtorture_test && \
+              /workspace/tools/smbtorture/regen.sh --report-only \
+                  --build-dir /build \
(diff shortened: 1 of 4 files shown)


jrhftw (commented):
Attaching ChatGPT code review here.

[PR1500_Code_Review.md](https://github.com/user-attachments/files/30718828/PR1500_Code_Review.md)

pdunlap:
> Attaching ChatGPT code review here.
> 
> [PR1500_Code_Review.md](https://github.com/user-attachments/files/30718828/PR1500_Code_Review.md)


Code is updated, responses including a couple things deferred to the future:
[FOLLOWUPS.md](https://github.com/user-attachments/files/30722276/FOLLOWUPS.md)

pdunlap:
Setting to draft while I work through diagnosing a nightly build failure. OK, good now, I had to tweak regen.sh to adjust permissions.

benjarvis (approved):

PR #1511, opened 2026-08-11
SMB and NFS fixes

ilya-nasuni:


benjarvis:
Thanks all!   I pulled out the NFSv4 "/" export fix and rewrote it in a stand alone PR here: https://github.com/chimera-nas/chimera/pull/1520

PR #1514, opened 2026-08-12
diskfs: gate by-fh loads on home-block allocation (close reclaim stale-load window)

benjarvis:
## Problem

A by-fh resolution that misses the inode cache faults the dinode straight from
its home block, gated only by `sm_inum_valid()` — a pure address-range check.
That leaves a window after a remove:

- the resident `nlink==0` struct is dropped when the retire transaction commits
  (`durable_wm`), but
- the tombstone dinode is not pushed home to the raw block until a later
  checkpoint/trim.

In that gap a fresh fault reads the pre-tombstone `nlink==1` from the home block
and spuriously resolves a removed handle as live. This is diskfs-specific
(memfs/cairn free synchronously); it surfaces as intermittently stale file
handles resolving instead of returning `NOENT`/`STALE`.

## Fix

Gate the cold-load path on **home-block allocation** rather than mere
addressability. `sm_inum_allocated()` adds an in-memory free-map floor query to
the existing range check. The home block is returned to its AG free pool at the
retire txn's *apply* (a few microseconds after commit, far ahead of the
push-home), so a reclaimed inum reads as gone from apply-time on. A block reused
for a new inode reads allocated again but fails the generation check below.

Because a dinode's home-block address is fixed by its inum (dinodes never
relocate), an allocated home block always means a live inode lives there, so the
check never false-negatives a live inode. The free tree tracks committed state
(the reservation allocator leaves an allocated-but-not-yet-durable block in the
tree until its txn retires), and a by-fh load only ever runs for an inode whose
create/remove is durable, so a live inode never reads as free.

## Locking

`ag->lock` is otherwise a leaf lock — the allocator and the intent-log apply
thread take it alone, and the allocator may take it and then evict an inode
(`ag -> shard`). A by-fh resolution reaches the gate holding the op's inode
locks, so a **blocking** acquire here would invert that order and can deadlock
the apply thread against the allocator. The gate therefore uses `trylock` and
falls back to the prior load on contention: never worse than before, and the
window it closes only matters right after a remove when the AG is quiescent, so
the trylock succeeds exactly when it counts.

Draft: found via the in-development NFS3 model-based test harness (not included
here); opening early for review of the approach.

### DIFF
--- a/src/vfs/diskfs/diskfs_inode.c
+++ b/src/vfs/diskfs/diskfs_inode.c
@@ -337,13 +337,25 @@ diskfs_inode_acquire(
     if (unlikely(!inode)) {
         /* Not resident: either evicted (its dinode is durably home -- eviction
          * only drops CLEAN inodes) or genuinely absent.  Fault it in from disk
-         * whenever the inum is within allocated space; the on-disk dinode read
-         * validates inum/gen/nlink and yields ENOENT if it isn't really there.
-         * (This must not gate on `mounted` -- a freshly-formatted FS evicts
-         * too, so a miss is not necessarily ENOENT.) */
+         * only when the inum's home block is still ALLOCATED; the on-disk dinode
+         * read then validates inum/gen/nlink and yields ENOENT if it isn't
+         * really there.
+         *
+         * The allocation check (not a bare range check) is what closes the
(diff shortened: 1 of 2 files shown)


benjarvis:
Solving this a better way, closing.

PR #1517, opened 2026-08-12
Smbtorture: browsable HTML report for matrix runs

pdunlap:
A full smbtorture matrix run leaves a results.txt (one PASS|FAIL|TIME|SKIP cell per backend/subtest) plus one smbt-fail.<backend>.<subtest>.log per non-passing cell — roughly 880 files and 8MB for the current catalog. The only consumer today is regen.py compare, which answers "did anything regress" in the nightly job summary but leaves no path from a named cell to the log that explains it. Reading the detail means grepping a directory.

tools/smbtorture/report.py renders one run as a single self-contained HTML file: backends → suites → failing tests → full log via nested <details>, so it needs no server, no JavaScript and no external assets. It reuses regen.py's parse_results / parse_cmake_lists / split_entries / parent_of / log_path rather than reimplementing the results and CMake parsing.

Every failure is classified against the check_ENABLED_<backend> minusSMBTORTURE_DISABLED_SUBTESTS): inside a backend's gated set is a regression, outside it a known gap. The page stamps which CMakeLists it judged against, that file's git revision, and a dirty marker — load-bearing, not decorative, since an old results tarball read against newer allowlists yields phantom regressions, and a stamped report identifies itself instead of being silently wrong.

Log rendering is deliberately additive: a headline is extracted for the summary line, but every line still reaches the page — torture output first, the chimera server log behind a collapsed section. The two most informative findings in
the run that motivated this were a client PANIC backtrace and mojibake inside a filename; neither survives asummarising extractor.

Build wiring. tools/ was outside the build entirely, so the report's unit tests could only be run by hand, against the project rule that all testing goes through ctest. A new tools/CMakeLists.txt registers the suite as chimera/tools/smbtorture/report, guarded on NOT DISABLE_TESTS like every other test subdirectory, with its own find_package(Python3) so it can't be silently dropped if the PyNFS block's call moves behind a feature guard. A genuinely absent Python3 is announced rather than skipped without a trace.

CI. The nightly matrix already produces everything the report consumes and uploads it as smbtorture-matrix-amd64, so rendering happens in the same run, inside the same container — publishing still depends on nothing outside the image. It runs last in the chain, after the newly-failing verdict is settled, so a rendering failure cannot skip the regression gate; report.py exits non-zero only when it could not read the results or write the page, which is the same contract the existing drift.md render runs under. safe.directory /workspace precedes it because the runner checks out the workspace while the container runs as root, and without it git refuses the repo for dubious ownership and the revision stamp degrades to "unknown revision" — exactly the unlabelled report the stamp exists to prevent. The step summary names the artifact rather than linking the page, since a summary cannot address a file inside an artifact.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -576,6 +576,23 @@ jobs:
       # matrix or a Samba upgrade changes the catalog, and then the live one is
       # the honest denominator.  regen.sh writes that file on every run, scoped
       # or not, so nothing here has to infer the scope from which files exist.
+      #
+      # report.py runs last, after the verdict flag is settled: it renders the
+      # same results into one self-contained HTML file, so the artifact carries
+      # a path from a named cell down to the log that explains it rather than
+      # ~880 loose smbt-fail.*.log files.  Last on purpose -- everything ahead
+      # of it in the chain either publishes or decides the exit status, and a
+      # rendering failure must not be able to skip the regression verdict.
+      # It exits non-zero only when it genuinely could not read the results or
+      # write the page (a regression is still a 0), so && is the same contract
+      # the drift.md render above already runs under.
+      #
+      # safe.directory: the page stamps the revision of the baseline it judged
+      # against, which is the whole defence against an old results set read
+      # against newer allowlists.  /workspace is checked out by the runner and
+      # the container is root, so without this git refuses the repo as dubious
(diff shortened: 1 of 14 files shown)


pdunlap:
Attached example html report: [report.html](https://github.com/user-attachments/files/30990822/report.html)
Report was harvested from [Nightly #91](https://github.com/chimera-nas/chimera/actions/runs/31619767896)

PR #1524, opened 2026-08-13
smb: return `SMB2_STATUS_INVALID_INFO_CLASS` for FileStreamInformation when streams unsupported

### LINKED ISSUE #1475: [SMB] FileStreamInformation returns STATUS_NOT_IMPLEMENTED when ADS unsupported, where MS-FSA specifies STATUS_INVALID_INFO_CLASS
**Severity:** low (high confidence) · **Effort:** ~1 SP (extra-small) · **Audit ref:** V3-112 (2026-06 server-side conformance audit, v3)

## Problem
A FileStreamInformation query against a share/backend without alternate-data-stream support returns STATUS_NOT_IMPLEMENTED (0xC0000002). MS-FSA specifies STATUS_INVALID_INFO_CLASS (0xC0000003) for object stores that do not support streams. Since `named_streams` defaults off and only memfs has the capability, this is the default-path response.

## Spec
MS-FSA 2.1.5.12.29: object stores that do not support alternate data streams SHOULD return STATUS_INVALID_INFO_CLASS.

## Code location
`src/server/smb/smb_proc_query_info.c:220-224`

## Evidence
`if (!config.named_streams || !(capabilities & CHIMERA_VFS_CAP_NAMED_STREAMS)) chimera_smb_complete_request(request, SMB2_STATUS_NOT_IMPLEMENTED)` returns 0xC0000002 instead of the spec's STATUS_INVALID_INFO_CLASS. Verified still present on main.

## Impact
Clients probing stream support receive the wrong NTSTATUS, which can mislead capability detection; a more conformant alternative is to synthesize a single default `::$DATA` entry from the file size.

## Dedup
No open issue covers this.

---
<sub>Filed from the 2026-06 server-side NFS/SMB/S3 conformance audit (v3, cross-checked against upstream/main @ 630c093e). Independently verified against current main.</sub>

sgon-nasuni:
## Summary

When SMB named streams are disabled, or when the selected backend does not support named streams, Chimera previously returned `STATUS_NOT_IMPLEMENTED`.

MS-FSA 2.1.5.12.29 specifies that object stores without alternate data stream support should return `STATUS_INVALID_INFO_CLASS`.

## Change

- Return `SMB2_STATUS_INVALID_INFO_CLASS` instead of `SMB2_STATUS_NOT_IMPLEMENTED` from the unsupported `FileStreamInformation` path.

## Fixes

Fixes #1475

### DIFF
--- a/src/server/smb/smb2.h
+++ b/src/server/smb/smb2.h
@@ -1183,6 +1183,16 @@ typedef uint8_t smb2_guid[SMB2_GUID_SIZE];
  */
 #define SMB2_FILE_ALL_INFO_FIXED_SIZE               100
 
+/*
+ * FileStreamInformation's per-entry fixed portion is NextEntryOffset(4) +
+ * StreamNameLength(4) + StreamSize(8) + StreamAllocationSize(8) = 24 bytes,
+ * followed by the variable-length StreamName.  A client must nonetheless offer
+ * at least 32 bytes of OutputBufferLength before the level is answered at all:
+ * that is the minimum Windows enforces, what Samba's marshall_stream_info
+ * checks, and the value smbtorture's smb2.getinfo.qfile_buffercheck asserts.
+ */
+#define SMB2_FILE_STREAM_INFO_FIXED_SIZE            32
+
 
 struct smb2_file_directory_information {
     uint32_t next_entry_offset;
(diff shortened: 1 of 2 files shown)


benjarvis:
The change here causes smbtorture to fail because it iterates over every info class and tests them all. NOT_IMPLEMENTED it accepts and moves on, INVALID_INFO_CLASS it flags as an error.   I guess it's debatable if the test is correct to do that or not, but either way this needs a little thought.  Windows and Samba evidently both synthesize a ::$DATA stream in this scenario, so I've added code along those lines here.

PR #1548, opened 2026-08-20
nfs: stop the DRCs replaying one client's reply to another

arenaud16:
The NFSv4.0 duplicate-request cache caches every minorversion-0 COMPOUND,
read-only ones included, and keys it by `{client address with the ephemeral
port stripped, xid, request checksum}`. Nothing in that key separates two
connections from one host, so a byte-identical COMPOUND at the same xid is
answered from the cache no matter who sent it. Two consequences:

- a read-only COMPOUND (`PUTROOTFH+LOOKUP`, `PUTFH+READDIR`) is served an
  arbitrarily old reply, and
- a `CREATE` from a second, concurrently live connection is answered `NFS4_OK`
  from the cache without ever executing, so a real mutation is acknowledged and
  discarded.

Any client that restarts its xid at a fixed value collides with its own earlier
runs and with sibling clients on the same host; a fresh pynfs `NFS4Client`
starts at 0, which is how this surfaced.

This is a regression. The v4.0 reply cache added in cd259a97 keyed on the
connection (`entry->conn == conn && entry->xid == xid`), so neither symptom was
reachable. 601a8d03 replaced it with an instance of the NFSv3 connectionless
DRC, keyed "like NFSv3 by {client,xid,proc,cksum}" — dropping the connection
from the key is the defect. That change had a real goal (a client's reply band
following it to another instance on failover), so the fix below keeps the
address-keyed band rather than simply reinstating a per-connection cache.

## Changes

- **Install the v4.0 cache only when `server.nfs4_drc` is set**, as the NFSv3
  DRC already is with `server.nfs3_drc`. It previously installed
  unconditionally with the flag gating only KV persistence, so a server that
  had opted out still carried the cache.

- **Cache a 4.0 COMPOUND only when it carries an operation that must not be
  re-executed.** The operations are still XDR at dispatch, so the filter gates
  the insert rather than the lookup: `chimera_nfs4_compound` cancels the armed
  reply capture unless the compound is cacheable. Never inserted is also never
  matched, because the key covers a checksum of the request. WRITE is excluded,
  as on the v3 path, and so are the seqid-sequenced state operations — RFC 7530
  §9.1.7 gives those a per-owner reply cache, which the RFC calls the more
  reliable of the two.

- **Record which connection a cached reply was captured on.** An entry replays
  to that connection, or to any connection once that one is gone. A band loaded
  from the KV store is claimed by the connection whose request triggered the
  load, so the persistent path gets the same protection as the captured one.

- **Bound entry age at 120s.** Entries previously lived until FIFO eviction at
  the 4MB cap, so a reply could be replayed arbitrarily long after capture.

- **Move the NFSv4.0 reply band from KV record type `0x08` to `0x09`.** Records
  in `0x08` were written by builds that cached read-only compounds too, and
  nothing in a record distinguishes the two, so the band is abandoned rather
  than hydrated into a cache that would replay stale LOOKUP and READDIR replies
  after an upgrade.

- **Compute the minorversion peek offset in 64 bits.** `tag_len` is read
  straight off the wire before the request is decoded, and in 32-bit arithmetic
  `4 + tag_len + pad` wraps: `tag_len` `0xFFFFFFF5`–`0xFFFFFFF8` all land on
  `off == 0xFFFFFFFC`, whose `off + 4` wraps to 0 and passes the bounds check,
  reading roughly 4GB past the buffer. A crafted COMPOUND tag could therefore
  crash the server.

The connection-origin guard and the age bound live in the shared core, so the
NFSv3 cache gets them too — it has the same key and the same exposure.

## Notes for review

- The connection origin is deliberately entry *payload* rather than part of the
  hash key. Keying on it would give each connection its own entry, but a
  reconnecting client would then miss its own record entirely, and answering a
  retransmit that arrives on a new connection is the whole point of an
  address-keyed cache. The cost is that two live connections at one address
  issuing byte-identical requests at the same xid share one entry, so the later
  capture takes ownership. That direction is safe: it degrades to re-executing
  a retransmit, never to answering one connection with another's reply.

- pynfs RPLY14 replays a CREATE at the same xid and requires the cached reply,
  so the 4.0 runs of the pynfs wrapper now enable `nfs4_drc`; 4.1+ replay goes
  through the session reply cache and is unaffected.

- `kvm_nfs40_drc_reboot_test_wrapper.sh`'s hydration assertion moves with the
  band to type `0x09`. The KVM legs have not been run against this branch — the
  container used here does not register them — so they are worth a run before
  merge, as they are the only coverage for the band change.

### DIFF
--- a/kvm/kvm_nfs40_drc_reboot_test_wrapper.sh
+++ b/kvm/kvm_nfs40_drc_reboot_test_wrapper.sh
@@ -209,8 +209,8 @@ if [ "${RC2:-1}" != "0" ]; then
     FAIL=1
 fi
 
-if grep -q "conn DRC (type 0x08): hydrated" "$CHIMERA_LOG"; then
-    RELOADED=$(grep -oP 'conn DRC \(type 0x08\): hydrated \K[0-9]+' "$CHIMERA_LOG" | tail -1)
+if grep -q "conn DRC (type 0x09): hydrated" "$CHIMERA_LOG"; then
+    RELOADED=$(grep -oP 'conn DRC \(type 0x09\): hydrated \K[0-9]+' "$CHIMERA_LOG" | tail -1)
     echo "=== chimera B hydrated ${RELOADED} NFSv4.0 reply record(s) for the returning client ==="
     if [ "${RELOADED:-0}" -lt 1 ]; then
         echo "FAIL: no NFSv4.0 reply records hydrated for the returning client"
--- a/scripts/nfs4_v40_drc_replay_test_wrapper.sh
+++ b/scripts/nfs4_v40_drc_replay_test_wrapper.sh
@@ -1,0 +1,155 @@
+#!/bin/bash
+# SPDX-FileCopyrightText: 2026 Chimera-NAS Project Contributors
+#
+# SPDX-License-Identifier: Unlicense
+
+# Usage: nfs4_v40_drc_replay_test_wrapper.sh <chimera_binary> <pynfs_dir> <test_script>
+#
+# Cross-client replay regression: stands up a chimera NFS server with the NFSv4.0
+# duplicate-request cache ENABLED (server.nfs4_drc) in a network namespace, then
+# drives the replay test against it.  The cache is opt-in, so a run without the
+# flag would prove nothing -- the code under test would not be installed.
+#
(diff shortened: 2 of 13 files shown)


arenaud16:
Closing I do not like this fix which seems to be a bandaid for a problem introduce in 601a8d03

PR #1550, opened 2026-08-21
nfs: destroy the NFSv4.1 session when the last mount goes

benjarvis:
The NFSv4.1 client established a session and then simply stopped talking. It
never sent `DESTROY_SESSION` or `DESTROY_CLIENTID`, so to a server a clean
shutdown was indistinguishable from a client that had crashed: it kept the
clientid's state — and the open handles behind it — until the lease expired.

For that whole window the export cannot be unmounted (`nfs4_lease_time`
defaults to 90s), and a client that restarts strands a fresh clientid on every
cycle.

## Following the kernel

The Linux client ties this to the lifetime of its `nfs_client`: mounts of a
server share one, and the last put drains the session, sends
`DESTROY_SESSION` and `DESTROY_CLIENTID`, and only then tears down the
transport. This does the same, on the last umount of a server.

That point is chosen for two reasons. It is where the last reference actually
drops, and it is the only place where both halves of the exchange are still
possible — the server's transport is up and an ordinary event loop is running
to carry the replies, so the teardown chains through completion callbacks like
any other operation. Module teardown is too late: the per-thread rpc2
transports are already gone, and nothing is left to speak on.

## The slot pool, which is why this wasn't done before

The old comment in `chimera_nfs4_umount` explained that unpublishing a session
while per-thread slot tables still held ids out of it would corrupt them.
Linux doesn't have that problem because its slot table lives in the session and
dies with it. Here the tables are per (thread, server) and deliberately
lock-free, so rather than moving them, each one is given the identity of the
session its ids came from:

- the session is **refcounted** — one reference for the server that published
  it, one per slot table holding its ids — so the pool stays mapped until the
  last table lets go;
- each table **records the session it was built against**, and the acquire path
  compares that against the server's currently published session. When they
  differ, the session was torn down and replaced by a remount, so the table
  returns its ids to the old pool and rebuilds against the new one.

Comparing rather than dereferencing is what makes that safe, and a table is
only ever repaired by the thread that owns it, so nothing reaches across
threads into lock-free state. The hot-path cost is one pointer compare.

## Also fixed

`chimera_nfs4_slot_table_destroy()` returned ids to `server->nfs4_session`
rather than to the session they were leased from. Those were the same pointer
only because a session was never replaced; with teardown they are not, and the
ids would have gone back to the wrong pool.

## Releasing the opens first

Destroying the session while opens remain is legal — state is clientid-scoped,
not session-scoped, so `DESTROY_SESSION` carries no "must have no state"
precondition and chimera's server correctly doesn't impose one. But it is not
*right*: it strands the opens with no session through which the client could
still `CLOSE` them. Linux never has to think about this because its VFS refuses
to unmount a filesystem with open files.

Chimera's umount drain (#1546) gives the same guarantee — it purges every handle
on the mount and issues a `CLOSE` for each before dispatching the backend
`UMOUNT`. The problem was one layer down: `chimera_nfs4_close` freed its local
state and reported success without sending anything. The wire `CLOSE` was a
TODO, so every close was a no-op on the server, which is exactly why
`DESTROY_CLIENTID` answered `CLIENTID_BUSY`.

Sending one per handle is not safe. The server keys open state on (open-owner,
file handle) and this client uses one open-owner per server, so a second `OPEN`
of an already-open file is an **upgrade** — the same stateid back with a bumped
seqid — not an independent open. One `CLOSE` releases the file for every opener,
so a per-handle `CLOSE` destroys state the others are still using. `fstest`
reproduces it: a re-open overlapping the previous handle's close leaves the
write on the surviving handle rejected, two handles holding one stateid that
differ only in seqid.

So the file's open on the server is counted separately from the handles that
have it open — a small table keyed on the wire handle holding a refcount and the
stateid, with `CLOSE` sent only when the last handle lets go.

### Why counted apart, rather than sharing the open state

Because the pNFS layout lives in the open state, and a layout is per handle
here. `nfs4_write.c` routes a write from the handle's own state: a VALID layout
sends it to the data server, and the MDS then learns the size only from a
LAYOUTCOMMIT this proxy issues lazily. A handle *without* a layout writes to the
MDS and the size follows for free.

Share one layout between two handles and every write goes to the DS, leaving
correctness to depend entirely on that lazy LAYOUTCOMMIT — and a recall (a
truncate from another client is enough) then fences the layout with the writes
still unpublished, so a reader falls back on an MDS that never heard about them.
That is what `pnfs/proxy_recall_truncate` reads back as `differ: char 1`, and it
is why an earlier revision of this PR failed `kvm-pnfs` on every kernel and
backend. Keeping the count separate leaves the layout lifetime exactly as it
was.

## Two things reviewers should know

**A correction to an earlier revision of this description.** It previously said
this change does not fix the "cannot remove a filesystem after unmounting an
NFSv4 share" symptom, and that whatever pinned the mount was therefore not
clientid-scoped state. That was wrong. It was clientid-scoped state: the opens
this client never released. Releasing them is what the last commit does.

### DIFF
--- a/src/vfs/nfs/CMakeLists.txt
+++ b/src/vfs/nfs/CMakeLists.txt
@@ -41,6 +41,7 @@ add_library(chimera_vfs_nfs SHARED
     nfs4_mknod_at.c
     nfs4_open_at.c
     nfs4_open_fh.c
+    nfs4_open_state.c
     nfs4_pnfs.c
     nfs4_read.c
     nfs4_slot.c
--- a/src/vfs/nfs/nfs.c
+++ b/src/vfs/nfs/nfs.c
@@ -127,6 +127,9 @@ chimera_nfs_destroy(void *private_data)
 
     for (i = 0; i < shared->max_servers; i++) {
         if (shared->servers[i]) {
+            chimera_nfs4_open_file_drain(shared->servers[i]);
+            pthread_mutex_destroy(&shared->servers[i]->open_state_lock);
+
             /* Release the server's reference on any session still published.
              * Reaching here with one means the module is going away with mounts
              * still up, so it was never destroyed on the wire; the memory goes
--- a/src/vfs/nfs/nfs.c
+++ b/src/vfs/nfs/nfs.c
@@ -127,11 +127,13 @@ chimera_nfs_destroy(void *private_data)
 
     for (i = 0; i < shared->max_servers; i++) {
         if (shared->servers[i]) {
-            /* Free NFS4 session if present */
+            /* Release the server's reference on any session still published.
+             * Reaching here with one means the module is going away with mounts
+             * still up, so it was never destroyed on the wire; the memory goes
+             * once the last slot table lets go. */
             if (shared->servers[i]->nfs4_session) {
-                chimera_nfs4_session_pool_destroy(shared->servers[i]->nfs4_session);
-                pthread_mutex_destroy(&shared->servers[i]->nfs4_session->lock);
-                free(shared->servers[i]->nfs4_session);
+                chimera_nfs4_session_put(shared->servers[i]->nfs4_session);
+                shared->servers[i]->nfs4_session = NULL;
             }
             free(shared->servers[i]);
         }
--- a/src/vfs/nfs/nfs3_mount.c
+++ b/src/vfs/nfs/nfs3_mount.c
@@ -561,6 +561,8 @@ chimera_nfs3_mount(
 
         server->index = idx;
 
+        pthread_mutex_init(&server->open_state_lock, NULL);
+
         need_discover = 1;
 
         DL_APPEND(server->pending_mounts, request);
--- a/src/vfs/nfs/nfs4_cb.c
+++ b/src/vfs/nfs/nfs4_cb.c
@@ -409,6 +409,10 @@ chimera_nfs4_cb_exchange_id_callback(
     pthread_mutex_init(&session->lock, NULL);
     session->clientid = eid_res->eir_clientid;
 
+    /* The server's own reference, released when the last mount goes away and
+     * the session is destroyed on the wire (chimera_nfs4_umount). */
+    atomic_store(&session->refcnt, 1);
+
     /* The MDS confirms pNFS support by echoing USE_PNFS_MDS; only then does the
      * client issue LAYOUTGET.  Otherwise it transparently stays non-pNFS. */
     server->mds_pnfs_capable = server->pnfs_requested &&
--- a/src/vfs/nfs/nfs4_close.c
+++ b/src/vfs/nfs/nfs4_close.c
@@ -6,6 +6,163 @@
 #include "nfs4_open_state.h"
 #include "nfs4_pnfs.h"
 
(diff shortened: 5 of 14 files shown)


benjarvis:
Moving to draft: the teardown is skipped in the common case.

`chimera_nfs4_umount` looks up `thread->server_threads[server->index]` to find a connection to send on, but that array is populated lazily per thread. The thread that handles the umount is often not one that has talked to this server, so `server_thread` is NULL and the code falls through to `chimera_nfs4_session_put()` without sending anything.

Instrumented on the nfs4 posix tests:

```
PROBE: umount refcnt=0 session=0xe2e6ab8a69d0
PROBE: server_thread=(nil) nfs_conn=(nil)
```

So the last reference does drop and the session is still published — the two hard parts work — but the goodbye is silently not sent. The refcount/slot-table rebuild half of the change is unaffected.

The fix is to obtain a connection rather than assume one: either connect on the umounting thread the way the mount path does, or hand the teardown to a thread that already has one. Will update.

benjarvis:
Pushed the connect-on-demand fix — the teardown now reaches the server, and the replies answer the question.

```
DESTROY_SESSION  rpc=0 nfs=10055  NFS4ERR_CONN_NOT_BOUND_TO_SESSION
DESTROY_CLIENTID rpc=0 nfs=10074  NFS4ERR_CLIENTID_BUSY
```

Getting a connection on demand was necessary but not sufficient. A freshly created connection has never carried a SEQUENCE, so it is not bound to the session, and NFSv4.1 refuses `DESTROY_SESSION` on it. `CLIENTID_BUSY` then follows correctly — the session still exists, so the clientid is legitimately busy.

Remaining work: run the teardown on a connection already bound to the session, or bind one first (`BIND_CONN_TO_SESSION`). Linux sidesteps this by destroying the session on the connection that was carrying it.

Note this also means `CLIENTID_BUSY` here does **not** yet tell us anything about lingering open state — it is explained entirely by the surviving session. The question of what pins the open handle after an NFSv4 unmount is still open, and gets a real answer only once `DESTROY_SESSION` succeeds.

benjarvis:
Added BIND_CONN_TO_SESSION. The session is now genuinely destroyed on the wire:

```
BIND_CONN_TO_SESSION rpc=0 nfs=0      OK
DESTROY_SESSION      rpc=0 nfs=0      OK
DESTROY_CLIENTID     rpc=0 nfs=10074  NFS4ERR_CLIENTID_BUSY
```

The first two are the goal of this PR and they now work.

The third is the interesting one. Last time `CLIENTID_BUSY` was fully explained by the surviving session, so it said nothing. Now the session is gone, and per RFC 5661 the only remaining reason for it is that the clientid still holds state — opens, locks or delegations.

**So this client leaves open state behind on the server when it unmounts.** That is what pins the server-side open handle (`cache_id=2`, `opencnt=1`) and makes the export un-unmountable afterwards, which is the symptom that started this whole line of work. It is a separate defect from session teardown, and worth its own investigation: either a CLOSE is not being sent for every open, or the server is not releasing on the ones that are.

The teardown now does its half correctly and surfaces the leftover state instead of hiding it behind a lease timeout.

PR #1561, opened 2026-08-24
smb: allow default data fork (::) open when named streams disabled

### LINKED ISSUE #1476: [SMB] With named streams disabled, default data-fork form 'file::$DATA' is rejected STATUS_OBJECT_NAME_INVALID instead of opening the base file
**Severity:** low (high confidence) · **Effort:** ~2 SP (small) · **Audit ref:** V3-113 (2026-06 server-side conformance audit, v3)

## Problem
With `named_streams` disabled, any CREATE whose name contains a ':' is rejected with STATUS_OBJECT_NAME_INVALID before stream parsing. This also rejects `name::$DATA`, which denotes the default unnamed data stream and is equivalent to opening `name` itself. The default data fork is not an alternate stream and must always be openable.

## Spec
MS-FSCC 2.1.5.1.1 / MS-FSA: the default (unnamed) data stream `::$DATA` is not an alternate data stream and is always a valid open of the base file.

## Code location
`src/server/smb/smb_proc_create.c:4255-4261`

## Evidence
`if (name contains ':') { if (!config.named_streams) { chimera_smb_complete_request(request, SMB2_STATUS_OBJECT_NAME_INVALID); return; } }` rejects every ':' name before `parse_stream_name` (smb_proc_create.c:3748-3758) runs. With `named_streams` ON the parser correctly maps `name::$DATA` to the base file, so the default-fork form is broken only when ADS is disabled. Verified still present on main.

## Impact
Applications/clients that open files using the explicit `file::$DATA` form fail on stream-disabled shares (the common default), even though they are merely opening the regular file.

## Dedup
No open issue covers this.

---
<sub>Filed from the 2026-06 server-side NFS/SMB/S3 conformance audit (v3, cross-checked against upstream/main @ 630c093e). Independently verified against current main.</sub>

sgon-nasuni:
## Summary

According to MS-FSCC 2.1.5.1.1, `file::$DATA` refers to the file's default data stream, not an alternate data stream.

When named streams were disabled, Chimera rejected this path with `OBJECT_NAME_INVALID`. It only checked for a colon and rejected the request before parsing whether the stream name was actually empty.

## Reproduction

```bash
smbclient //host/share -N -c "put file.txt readme.txt"
smbclient //host/share -N -c "get readme.txt::\$DATA out.txt"
# NT_STATUS_OBJECT_NAME_INVALID when named_streams is disabled (the default)
```

## Change

`chimera_smb_create()` now always parses the stream syntax first. When named streams are disabled, it rejects only non-empty stream names.

The empty-name form, `file::$DATA`, is treated as a normal file open regardless of the configuration.

Fixes https://github.com/chimera-nas/chimera/issues/1476

### DIFF
--- a/src/server/smb/smb_proc_create.c
+++ b/src/server/smb/smb_proc_create.c
@@ -4883,38 +4883,18 @@ chimera_smb_create(struct chimera_smb_request *request)
         return;
     }
 
-    /* Handle stream-name syntax (file:stream[:$DATA]).  When named streams are
-     * disabled (config off, or the backend lacks the capability) the legacy
-     * behavior is preserved: reject any ':' name with OBJECT_NAME_INVALID.
-     * Done here rather than in parse so the response is returned to the client
-     * instead of triggering a parse-error disconnect. */
+    /* Handle stream-name syntax (file:stream[:$DATA]).  "file::$DATA" (empty
+     * stream name, $DATA type) names the object's DEFAULT data fork, not an
+     * alternate stream (MS-FSCC 2.1.5.1.1), so it must open the base file even
+     * when named streams are disabled; the legacy behavior of rejecting any
+     * other ':' name with OBJECT_NAME_INVALID when named streams are disabled
+     * (config off, or the backend lacks the capability) is unchanged.  Parsing
+     * runs unconditionally so that distinction (explicit_data_fork) is known
+     * before the named_streams check.  Done here rather than in parse so the
+     * response is returned to the client instead of triggering a parse-error
+     * disconnect. */
     if (request->create.name_len > 0 &&
         memchr(request->create.name, ':', request->create.name_len)) {
-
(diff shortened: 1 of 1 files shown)


sgon-nasuni:
`::$DATA` refers to the file's default data stream, not an alternate stream. Opening `filename::$DATA` with `FILE_OPEN` when the file does not exist should return `OBJECT_NAME_NOT_FOUND`, just like opening `filename` directly. The spec does not appear to say that the request should be rejected simply because the path contains a colon.

However, Samba's `create_no_streams.no_stream` test expects `OBJECT_NAME_INVALID` in this case. This may be specific to Samba's implementation when the `streams_xattr` module is not enabled, rather than the expected Windows behavior.

My understanding is that `OBJECT_NAME_NOT_FOUND` is the spec-compliant result, while the Samba expectation is an implementation difference. Could someone confirm before I proceed with that assumption?

Issue #5, opened 2025-06-01
Implement S3 request signing checks


benjarvis:
S3 signing is currently ignored, need to add support for checking and a credential cache so we don't need to recalculate on every request.

benjarvis:
Resolved by PR #111

PR #1566, opened 2026-08-27
nfsaux: model-based tests for portmap, MOUNT, NLM and NSM (+3 server fixes)

benjarvis:
Consumes the new `nfsaux` corpus from chimera-nas/specs#5 and replays it against
an in-process chimera server, the way the NFSv3 and NFSv4 suites already do —
covering the four RPC protocols that surround NFSv3: portmap/rpcbind (100000
v2/v3/v4), MOUNT (100005 v3), NLM (100021 v4) and NSM (100024 v1).

Rebased onto main; `ext/specs` points at specs `main` (`49eca48`) — the
`9829b7c` main already carried, plus chimera-nas/specs#5, which is merged.

### Three server fixes, first two commits

Every one is reachable from the wire by an ordinary conforming client.

1. **`nfs: wrap the portmap DUMP results so a client can decode them`** — for a
   procedure returning a bare recursive type, xdrzcc emits an asymmetric codec:
   the marshaller writes the RFC 1833 chain, the generated *client*
   unmarshaller reads a single bare body and consumes no booleans, so every
   reply fails to decode. Wrapping the chain in a one-field struct makes the two
   agree; the wire encoding is byte-for-byte unchanged. `mount.x` already
   carries the same wrapper for `MOUNTPROC3_DUMP`, added for exactly this
   reason.
2. **`nlm: fix three faults in the asynchronous half of the protocol`**
   - every `send_call_NLMPROC4_*_RES` passed a NULL reply callback; the
     generated dispatcher invokes it unconditionally, so the server took a
     SIGSEGV the moment a client acknowledged the call — which every conforming
     client does;
   - `nlm4_send_res()` had no arm for the `_MSG` procedures and a silent
     `default: rc = 0`, so a `LOCK_MSG` that failed on a stale handle or an
     allocation failure sent *nothing* and left the client waiting forever;
   - `CANCEL_MSG` acknowledged and replied GRANTED without cancelling, on a
     comment that stopped being true when blocking locks started queueing in the
     claim core — a client that gave up on a blocking LOCK left it queued, to be
     granted later against a range nobody wanted.

   Plus the grant engine's shutdown callback, which read the `nlm_grant_ctx`
   that `evpl_thread_function` actually hands it as the `nlm_granter` passed to
   `evpl_thread_create`, and aborted in `evpl_core_epoll_remove()`. Tearing down
   a server that had ever served a blocking lock crashed.

### The suite, third commit

Two ctests, neither binding a port, needing a netns, or spawning a daemon:

- `mbtaux/probe_memfs` — ground truth with no corpus involved: which procedures
  the server registers and the constants their replies carry. Neither is
  derivable from the RFCs, so pinning them here means a change on either side
  goes red at the source instead of as a mass divergence in the replay.
- `mbtaux/batch_memfs` — 42 traces replayed back to back, ~25 s.

`nfs3_mbt_common.h` grows the auxiliary services behind an opt-in flag, so the
existing suites pay nothing. The NLM connection also *exports* NLM_V4, because
the asynchronous half of that protocol is answered by the server calling
`NLMPROC4_*_RES` back on the same connection — a client that does not serve the
program cannot see its own results.

Isolation needed more than the NFSv3 suite's fresh-filesystem-per-trace: the
lock table, the client table, the monitor list and the NSM state number are all
server-global. Caller names are qualified per trace, every caller is `FREE_ALL`'d
at teardown (a leftover lock holds an open handle and `rmfs` would never
succeed), and the NSM state number — which only ever rises — is read at trace
start and every prediction shifted by it.

`--paranoid` is on for the batch. Comparing replies alone does not merely
*mislocate* a lock-table divergence, it can miss one outright: a lock the server
failed to grant or release is invisible until some later operation happens to
collide with it, and nothing guarantees one does. The flag probes each file's
ranges against the model's own lock set after every step, and it is what caught
the release-ordering behaviour the model now encodes (a client's locks are
released one at a time with a queue pump after each, so a narrow waiter
satisfied by a partial release beats a wider one).

### Coverage

Line coverage of the eight files these protocols live in, `ctest -L
"nfs3_mbt|nfs4_mbt"` before vs. the same plus `nfsaux_mbt`, one clang Coverage
build and one fixed object set:

| file | lines | before | after |
|---|---|---|---|
| `nfs_nlm.c` | 922 | 0.0% | 85.9% |
| `nfs_nsm.c` | 447 | 1.8% | 46.8% |
| `nfs_nlm_granted.c` | 282 | 1.4% | 76.2% |
| `nfs_mount.c` | 276 | 38.0% | 85.1% |
| `nfs_portmap.c` | 194 | 5.2% | 88.7% |
| `nfs_nlm_state.c` | 171 | 17.5% | 58.5% |
| `nfs_nsm_state.c` | 125 | 16.8% | 49.6% |
| `nfs_external_portmap.c` | 169 | 0.0% | 0.0% |
| **those eight** | 2,586 | 6.9% | **69.0%** |
| **src/server/nfs** | 27,058 | 45.0% | **57.2%** |

Measured against *every* NFS-labelled test rather than the quint suites alone,
744 of those 2,586 lines remain. Roughly half needs a second process (external
`rpcbind`, the reboot-notify worker, IPv6 peer addresses); ~150 needs a
persistent KV backend and a restart, which would also turn on the NLM grace
window — the model side of that already exists as the `nfsauxGrace` profile and
generates no traces only because no server configuration matches it yet.

### Verification

- `ctest -R "^chimera/server/nfs/"` — 20/20
- each commit builds on its own; the first two also pass the 18 pre-existing NFS
  tests without the suite present
- scan-build clean, uncrustify clean

### DIFF
--- a/src/nfs_common/portmap.x
+++ b/src/nfs_common/portmap.x
@@ -1,5 +1,6 @@
 /*
  * SPDX-FileCopyrightText: IETF Trust and the persons identified
+ * SPDX-FileCopyrightText: 2026 Chimera-NAS Project Contributors
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
@@ -40,6 +41,23 @@ struct rpcb {
 struct rp__list {
     rpcb rpcb_map;
     rp__list *next;
+};
+
+/*
+ * RFC 1833 declares the rpcbind DUMP procedures as returning a bare
+ * "rp__list *" (an optional pointer chain).  xdrzcc emits an asymmetric
+ * codec for a procedure whose result is a recursive type used directly:
+ * the marshaller writes the RFC encoding (value-follows boolean + body,
+ * repeated, terminated by a zero boolean) but the generated *client*
+ * unmarshaller decodes a single bare body and consumes no booleans, so
+ * every reply fails to decode.  Wrapping the chain in a one-field struct
+ * makes the two symmetric; the wire encoding of an optional pointer inside
+ * a struct is byte-for-byte the RFC's list, so this is not a protocol
+ * change.  mount.x carries the same wrapper (mountdumpres) for
+ * MOUNTPROC3_DUMP, for the same reason.
+ */
+struct rpcbdumpres {
+    rp__list *maps;
 };
 
 /*
@@ -162,6 +180,11 @@ struct pmaplist {
     pmaplist *next;
 };
 
+/* The PMAP v2 counterpart of rpcbdumpres; see the note there. */
+struct pmapdumpres {
+    pmaplist *maps;
+};
+
 /* Arguments to callit */
 struct call_args {
     unsigned int prog;
@@ -183,7 +206,7 @@ program PORTMAP {
         bool PMAPPROC_SET(mapping) = 1;
         bool PMAPPROC_UNSET(mapping) = 2;
         unsigned int PMAPPROC_GETPORT(mapping) = 3;
-        pmaplist PMAPPROC_DUMP(void) = 4;
+        pmapdumpres PMAPPROC_DUMP(void) = 4;
         call_result PMAPPROC_CALLIT(call_args) = 5;
     } = 2;
 
@@ -191,7 +214,7 @@ program PORTMAP {
         bool rpcbproc_set(rpcb) = 1;
         bool rpcbproc_unset(rpcb) = 2;
         string rpcbproc_getaddr(rpcb) = 3;
-        rp__list *rpcbproc_dump(void) = 4;
+        rpcbdumpres rpcbproc_dump(void) = 4;
         rpcb_rmtcallres rpcbproc_callit(rpcb_rmtcallargs) = 5;
         unsigned int rpcbproc_gettime(void) = 6;
         netbuf rpcbproc_uaddr2taddr(string) = 7;
@@ -202,7 +225,7 @@ program PORTMAP {
         bool RPCBPROC_SET(rpcb) = 1;
         bool RPCBPROC_UNSET(rpcb) = 2;
         string RPCBPROC_GETADDR(rpcb) = 3;
-        rp__list *RPCBPROC_DUMP(void) = 4;
+        rpcbdumpres RPCBPROC_DUMP(void) = 4;
 
         /*
          * NOTE: RPCBPROC_BCAST has the same functionality as CALLIT.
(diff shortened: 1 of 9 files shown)


benjarvis:
Parked on #913.

chimera-nas/specs#5 is merged, so the submodule pointer here needs to move to
specs `main` (`49eca48`). But specs `main` also carries chimera-nas/specs#4
(`nfs4-rfc7862-data-ops`), whose chimera side is #913 — and bumping to `49eca48`
before #913 lands takes four pre-existing tests red:

- `mbt4/batch_memfs` aborts with `no encoder for RClone`: specs#4 marks CLONE,
  READ_PLUS, IO_ADVISE and WRITE_SAME `FeatMandatory` for the memfs profiles,
  and #913 is what implements them (and teaches `nfs4_mbt_replay.c` to encode
  them).
- `mbt/batch_{memfs,diskfs,cairn}` diverge on `OCreate`: specs#4 re-based the
  NFS3 CREATE model on RFC 1813 rather than probed behaviour (UNCHECKED over a
  directory → ISDIR, over a symlink/fifo/socket → EXIST, search permission
  taking precedence), and chimera answers OK.

Pinning at this PR's own specs commit instead is not a way around it: specs
publishes bundles only on push to `main`, keyed by the merge SHA, and chimera CI
has `oras` but no quint — so `SPECS_BUNDLE_AVAILABLE` would go OFF and every MBT
replay ctest, including the one this PR adds, would silently *unregister* rather
than run.

The nfsaux suite itself passes against both corpora; nothing here is waiting on
its own review. Once #913 merges I'll rebase onto main, move the pointer to
`49eca48`, and re-verify.

benjarvis:
Unparked — #913 is in, so this is rebased onto main (`f37a8f31`) and the
submodule now points at specs `main` (`49eca48`), which is the `9829b7c` main
already carried plus chimera-nas/specs#5.

Both halves of the earlier blockage were resolved on main rather than papered
over, for the record:

- the four NFS4.2 data ops are implemented and `nfs4_mbt_replay.c` can encode
  them (#913), so `mbt4/batch_memfs` no longer aborts on `RClone`;
- `472fcf44` fixed the server to reject NFS3 UNCHECKED create over a
  non-regular name, and `9c0a68ee` registered the residual search-perm and
  exclusive-retry deviations.

Verified on the rebased tree: `ctest -R "^chimera/server/nfs/"` 20/20,
uncrustify clean. Nothing in the three commits changed.

Issue #8, opened 2025-06-01
Add data cache to the VFS


benjarvis:
The VFS caches metadata but not data.  Data is the harder part.   Can look to the linux kernel for inspiration.

PR #1569, opened 2026-08-28
nfs4: model-based pNFS tests against an in-process MDS and data servers

benjarvis:
Extends the NFSv4 quint suite to pNFS, on memfs and on diskfs, and fixes the six defects replaying it found.

Depends on **chimera-nas/specs#8** (the `nfs4pnfs` corpus). chimera-nas/libevpl#141 (a poll-handle use-after-free) has merged, and the last commit advances `ext/libevpl` to the main that carries it; the `ext/specs` pin still points at that PR's branch, so re-point it at the merged commit before this lands.

## The harness

`nfs4_mbt_replay --pnfs N` stands a whole metadata-server cluster up inside the test process: the server it already ran becomes the MDS, plus N more `chimera_server` instances as flex-files data servers. Each runs in `data_server` mode, binding only the NFSv4 service — no mountd/portmap/NLM — so they share the process on their own inproc service names, and the MDS reaches them through the `nfs` client module, which inherits the inproc flavor at mount time.

So a LAYOUTGET drives the real steer → backing-file-create → blob-persist path with **no ports, network namespaces, daemons or VMs**, unlike the KVM pNFS matrix, and every trace still runs fully parallel on any host. Two data servers, so round-robin steering and both arms of the flex-files device encoder (an NFSv3 DS and an NFSv4.1 one) are covered. `--backend diskfs` puts the MDS on diskfs.

Two ctests, `chimera/server/nfs/mbt4/batch_pnfs_{memfs,diskfs}` (label `pnfs`), 24 traces each, ~5s and ~9s.

## Coverage

`ctest -L quint`, line coverage, without → with the two new batches:

| file | before | after |
|---|---|---|
| `src/server/nfs/nfs4_pnfs.c` | 1.3% | **47.4%** |
| `src/server/nfs/nfs4_layout_table.c` | 16.0% | **89.0%** |
| `src/server/nfs/nfs4_cb.c` | 8.4% | **86.3%** |
| `src/vfs/vfs_pnfs.c` | 17.6% | **85.9%** |
| `src/vfs/nfs/nfs4_mount.c` | 0% | **70.0%** |
| `src/server/nfs/nfs4_callback.c` | 30.0% | **40.3%** |

`nfs4_mount.c` is the MDS→DS control path — the `nfs` client module had no quint coverage at all before. The `chimera_vfs` component total goes 33.2% → 38.8%.

Still dark in `nfs4_pnfs.c`: the block/SCSI sourced-layout half (`lg_sourced_cb`, the block/scsi encoders, the devcache) and GETDEVICELIST/LAYOUTSTATS/LAYOUTERROR. Reachable by giving diskfs `block_layout: true` and asking for `LAYOUT4_BLOCK_VOLUME`, which the harness already parameterizes but no ctest drives yet.

## What it found

**`diskfs` never persisted the pNFS layout blob.** `diskfs_setattr_inode_cb` tested `set_attr->va_set_mask` for `CHIMERA_VFS_ATTR_PNFS_LAYOUT`, but `diskfs_apply_attrs()` has already rewritten that mask to what it applied, and it knows nothing about the blob. The bit was never set at the test, so a diskfs MDS persisted no layout while reporting success — then re-steered on every LAYOUTGET and handed the client a brand new, empty backing file, orphaning whatever it had written behind the previous one and leaking a file on a data server every time. The function already captures `orig_mask` for exactly this reason. memfs was unaffected.

**LAYOUTCOMMIT never checked the layout.** It applied the client's reported size with no look at whether the client held one — so any client could set any file's size, and *shrink* it, since `loca_last_write_offset` was applied whether or not it extended the file. Now: a stateid naming a layout this client no longer holds is BAD_STATEID (a stale seqid OLD_STATEID, as for LAYOUTRETURN), a read-only layout is BADLAYOUT, and the high-water mark extends the file but never truncates it. An RW LAYOUTGET over a layout first taken for READ now widens it, so that access is visible to the LAYOUTCOMMIT that needs it.

**4.1 LOCK minted a second lock stateid** for a lock-owner that already had one on the file. The existing-stateid lookup sat inside the `minorversion == 0` branch, because the rule it was written for is 4.0's BAD_SEQID answer. 4.1 retired that rule but not the identity rule it protected: one stateid per lock-owner per file, `other` stable for its life (RFC 8881 9.1.4.4).

**Two teardown leaks** a second in-process server makes visible: `vfs/nfs` freed only the head of the `shared->mounts` DL list, and `server/nfs` leaked the portmap per-proc metrics in data_server mode (a data server initializes the programs, then discards the server).

Plus, in libevpl (#141), **`evpl_remove_poll` invalidated other pollers' handles** by compacting the array — an intermittent heap-use-after-free tearing down any thread with two or more pollers not retired in reverse order. `batch_pnfs_diskfs` faulted in roughly half of all runs before that fix.

## Not fixed

The callback channel a client's first LAYOUTGET opens leaks at shutdown: a reference taken for an in-flight CB RPC outlives the client, so `nfs4_cb_path_teardown`'s unref never reaches zero. Once per surviving channel, only at exit, and only on a server that opens callback paths — so pNFS, or delegations, which the MBT harness leaves off. LSAN-suppressed with `src/server/nfs/tests/quint/suppressions.txt`, which explains it; a single-trace run is clean, only the batched replays reach it.

## Validation

`ctest -L quint` is 20/20 on this branch, memfs/diskfs/cairn alike. `make syntax-check` clean.

Note for CI: no specs bundle is published for a branch commit, so `SPECS_BUNDLE=AUTO` will fall back to building `ext/specs` in-tree (quint + python3, both in the dev image). That resolves itself when the specs pin moves to a merged main commit.

### DIFF
--- a/src/server/nfs/nfs.c
+++ b/src/server/nfs/nfs.c
@@ -682,6 +682,14 @@ nfs_server_destroy(void *data)
 
     if (shared->portmap_server) {
         evpl_rpc2_server_destroy(shared->portmap_server);
+    }
+
+    /* The portmap programs' per-proc metrics are allocated whenever an internal
+     * portmap was configured, which is not the same as ending up with a portmap
+     * *server*: a data server initializes the programs and then discards the
+     * server, because it binds only the NFS service.  Free on the condition
+     * they were allocated under, not on the server pointer. */
+    if (!chimera_server_config_get_external_portmap(shared->config)) {
         free(shared->portmap_v2.rpc2.metrics);
         free(shared->portmap_v3.rpc2.metrics);
         free(shared->portmap_v4.rpc2.metrics);
--- a/src/server/nfs/nfs4_pnfs.c
+++ b/src/server/nfs/nfs4_pnfs.c
@@ -600,6 +600,13 @@ ff_lg_emit(struct ff_layoutget_ctx *ctx)
 
     layout = nfs_layout_state_find(client, req->fh, req->fhlen);
     if (layout) {
+        /* One layout per (client, file), so a later RW LAYOUTGET over a
+         * layout first taken for READ widens that layout rather than making a
+         * second one -- otherwise the RW access the client just acquired is
+         * invisible to LAYOUTCOMMIT, which requires it. */
+        if (args->loga_iomode == LAYOUTIOMODE4_RW) {
+            layout->iomode = LAYOUTIOMODE4_RW;
+        }
         nfs_layout_state_bump(layout, client_short_id, &res->logr_resok4.logr_stateid);
     } else {
         nfs_layout_state_create(client, req->fh, req->fhlen, req->export_id, args->loga_iomode,
@@ -832,6 +839,10 @@ lg_sourced_cb(
     client_short_id = (uint32_t) client->client_id;
     layout          = nfs_layout_state_find(client, req->fh, req->fhlen);
     if (layout) {
+        /* See ff_lg_emit(): widen an existing READ layout to RW. */
+        if (args->loga_iomode == LAYOUTIOMODE4_RW) {
+            layout->iomode = LAYOUTIOMODE4_RW;
+        }
         nfs_layout_state_bump(layout, client_short_id, &res->logr_resok4.logr_stateid);
     } else {
         nfs_layout_state_create(client, req->fh, req->fhlen, req->export_id, args->loga_iomode,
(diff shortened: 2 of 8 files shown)


benjarvis:
### On the red `Analyze macOS Release (pre-merge)` check

Not from this PR. scan-build reports three findings, all in files this branch does not touch:

| file | finding | origin |
|---|---|---|
| `src/common/evpl_iovec_cursor.h:383` | `unix.cstring.UninitializedRead` | pre-existing |
| `src/vfs/smb/smb_ops.c:220` | `unix.cstring.UninitializedRead` | pre-existing |
| `src/server/nfs/tests/quint/nfs_aux_probe.c:759` | `unix.Malloc`, leak of `env.data_buf` | arrived with main's tip, b892822f |

(A fourth, `ext/libevpl/src/core/allocator.c:252`, is inside the `--exclude ext` path and is not counted in the three.)

The same job on an unrelated branch off the same main — `quint-s3`, run 33177663817 — fails with the identical three, which is what makes them main-wide rather than mine. `git diff --name-only main..HEAD` on this branch does not list any of the three files.

Leaving them alone rather than fixing them here: two are long-standing and the third belongs to the nfsaux work, and at least one other in-flight branch appears to be addressing them already, so a duplicate fix would just collide. A rebase picks the fix up once it lands on main.

Everything else on this PR is green, and `ctest -L quint` is 20/20 locally (memfs, diskfs and cairn).

PR #1582, opened 2026-08-30
vfs/claim: decide a cancel without waiting out the backend confirm

benjarvis:
`chimera_vfs_claim_cancel` could not answer while the service thread was confirming the very ticket being cancelled, so it spun on `sched_yield()` until that confirm's callback returned. The callback runs protocol code, and every caller of cancel is a teardown walking its own list under its own lock, so the wait closes a cycle:

- **Thread A** — NLM client reap (`nlm_client_release_all_locks`) holds `nlm_state.mutex` and spins in `chimera_vfs_claim_backend_ticket_cancel`.
- **Thread B** — the confirm it is waiting for, blocked on `pthread_mutex_lock(&shared->nlm_state.mutex)` inside `chimera_nfs_nlm4_lock_acquire_cb`.

Neither moves, and the spin burns a core, so it presents as a hang rather than a stall. This is the `chimera/server/nfs/mbtaux/batch_memfs` timeout at 180s that started appearing once #1574 made the quick tier run on every OS (run 33273945653, and the three hangs recorded in #1581). Reproduced locally under concurrency with the harness alarm disabled; the backtrace has both threads on the same ticket and the same claim state.

The reap path is not exotic — besides `FREE_ALL` and `SM_NOTIFY` it runs whenever a client's last connection drops (`nfs.c:856`), so any client disconnecting with a blocking lock outstanding can hit it.

## What changed

**Arbitrate instead of waiting.** An in-flight confirm is published as a core-owned op on `state->confirm_head` for exactly the window before its callback is invoked. A cancel that finds one claims it: the callback never fires, the caller owns the ticket, and the completion drops the record the backend granted. Because the op is the core's memory and carries a snapshot of what the completion needs, cancel can return the instant it decides and let its caller free the ticket — the completion never dereferences it again.

`false` now means *the callback owns the completion and may still be running*, rather than *the callback has returned*. Callers arbitrate with their own callback through their own lock rather than through this one.

**Only the deferred (service-lane) confirm is published.** An inline confirm runs under `chimera_vfs_claim_acquire` on the acquiring thread, where claiming the ticket would suppress a callback that thread is still waiting on and swallow the reply it owes. There cancel keeps reporting `false` exactly as before.

**The lane no longer assumes the backend answers inline.** It never had to: `chimera_vfs_dispatch` routes an op to a delegation thread whenever async delegation is configured (a supported, documented knob), and a clustered arbiter will never answer inline. With the old code that would have cleared the active marker before the callback fired — turning cancel's contract into a use-after-free — and let a queued token release overtake the confirm ahead of it. Dispatching a confirm now gates the drain until it completes; when the answer comes inline the gate clears before `project_range` returns, so the drain continues without a doorbell round trip.

**NLM (first commit).** When the cancel loses, the reaper used to leave the entry for the callback, which then took the GRANTED path and handed a lock to a client just told it holds none. Both sides already take `nlm_state.mutex`, so a flag is a complete hand-off. That branch is unreachable today — the wait means the callback has finished by the time the flag would be set — which is why it goes first: removing the wait is what makes it live.

## Notes for the reviewer

- **The deferred lane had no unit coverage at all.** The claim tests build a state with no vfs, which leaves the projecting path unreachable, so nothing exercised the FIFO or the cancel interaction. The new test forces the capability and drives both points at which a cancel can still claim a ticket, plus the point at which it must yield.
- **A separate, pre-existing hole in the same lane.** `nfstest_lock_memfs_nfs4.1` intermittently fails "Locking byte range on second process should be granted: got error EAGAIN" right after the holder unlocks — a stale backend record. `chimera_vfs_claim_backend_drain_releases` is supposed to close that by dispatching matching releases before a confirm, but it cannot find a release the service thread has already popped, and that release then races an inline confirm on another thread. This change does not fix it (it makes entries dwell longer in the FIFO where the drain can find them, if anything). Worth its own issue.
- **SMB's abort-vs-deferred-grant window is pre-existing and unchanged in kind.** `chimera_smb_lock_abort_parked` reads `resume_status` and unlinks from `lock_resume_head` on the `false` path, which assumes the callback finished. That assumption is already wrong on every backend except memfs, because a non-CAP_LEASE pump fires the callback directly and the wait never covered it. On memfs this narrows from "fully covered" to a few instructions; the durable fix is a flag under `lease_break_lock`, which both sides already take, and belongs with SMB lock teardown ownership rather than here.

## Verified

- All 30 in-process quint/MBT suites pass, including `mbtaux/batch_memfs`.
- 12 concurrent runs of the aux batch with no hang (it previously wedged within ~12).
- `vfs/claim_test`: 155 checks pass, including the 9 new ones.
- `smb/mbt/batch_memfs` and `vfs/zerorange_diskfs_io_uring` fail identically on clean `upstream/main` in this environment; pynfs and the KVM suites need privileges the local container lacks.

### DIFF
--- a/src/server/nfs/nfs_nlm.c
+++ b/src/server/nfs/nfs_nlm.c
@@ -429,10 +429,41 @@ chimera_nfs_nlm4_lock_acquire_cb(
     res.cookie.data = ctx->cookie.data;
 
     if (result == CHIMERA_CLAIM_GRANTED) {
+        bool reaped;
+
         pthread_mutex_lock(&shared->nlm_state.mutex);
-        entry->pending        = false;
-        entry->claim_inserted = true;
+        reaped = entry->reaped;
+        if (reaped) {
+            /* The client was reaped (FREE_ALL, SM_NOTIFY, or its last
+             * connection dropped) while this acquire was in flight, and the
+             * reaper could not claim the ticket, so it left the entry to us.
+             * Granting now would hand a lock to a client we have already
+             * told holds none -- drop it instead.  Still linked, so unlink
+             * under the same mutex the reaper used. */
+            DL_DELETE(ctx->client->locks, entry);
+        } else {
+            entry->pending        = false;
+            entry->claim_inserted = true;
+        }
         pthread_mutex_unlock(&shared->nlm_state.mutex);
+
+        if (reaped) {
+            chimera_nfs_debug(
+                "NLM LOCK: grant landed for reaped client '%s'; releasing",
+                ctx->client->hostname);
+            chimera_vfs_claim_release(vfs_state, entry->file_state,
+                                      &entry->claim);
+            chimera_vfs_state_put(vfs_state, entry->file_state);
+            entry->file_state = NULL;
+            chimera_vfs_release(thread->vfs_thread, entry->handle);
+            nlm_lock_entry_free(entry);
+            /* The original LOCK RPC was already answered with NLM4_BLOCKED
+             * (only blocked entries can still be pending here), so no reply
+             * is owed and no NLM_GRANTED callback is wanted. */
+            free(ctx);
+            return;
+        }
+
         res.stat = NLM4_GRANTED;
 
         /* Monitor the lock holder so we can SM_NOTIFY it to reclaim if we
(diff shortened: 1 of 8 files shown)


benjarvis:
Folded into #1574 as its last two commits: that PR is what makes the aux MBT batch run on every configuration, so the fix belongs with it rather than queued behind it.

Two pre-existing issues in the same lane are described in this PR's original description and are NOT fixed there — worth their own issues:

1. `nfstest_lock_memfs_nfs4.1` intermittently returns EAGAIN for a lock taken right after the holder unlocks. `chimera_vfs_claim_backend_drain_releases` cannot find a release the service thread has already popped, and that release then races an inline confirm on another thread.
2. `chimera_smb_lock_abort_parked` reads `resume_status` and unlinks from `lock_resume_head` on the `false` path, which assumes the callback finished. That is already wrong on every backend except memfs, since a non-CAP_LEASE pump fires the callback directly and the wait never covered it.

PR #1610, opened 2026-09-01
posix/quint: bring the SMB client proxy under the POSIX harness (was 0% covered)

benjarvis:
`src/vfs/smb` — the SMB2 **client** proxy VFS module — was at **0% coverage**:
0 of 129 functions, 0 of 2,800 lines. Nothing labelled `quint` executed it. Its
only test, `smb_mount_test`, is gated behind `CHIMERA_LIMITS_TESTING`. (This
predates the PR: #1609 reports the same 0% on the same day.)

This brings it under the POSIX model harness, the SMB2 sibling of the existing
`nfs3_`/`nfs4_` loopbacks: the posix client mounts an in-process chimera SMB
server through the smb proxy, over the inproc transport, backed by memfs.

## What lands in CI

`chimera/posix/smb/loopback_probe` — a corpus-free, green, quick-tier test over
that loopback. Measured over the probe alone:

| File | Lines |
|---|---|
| `smb_ntlm.c` | 0% → **86%** |
| `smb_ops.c` | 0% → **83%** |
| `smb_mount.c` | 0% → **81%** |
| `smb_umount.c` | 0% → **81%** |
| `smb_io.c` | 0% → **80%** |
| `smb.c` | 0% → **64%** |
| `smb_namespace.c` | 0% → **54%** |

It asserts two distinct kinds of thing, and says which is which:

- the operations the proxy **does** implement — create/write/read at depth,
  exclusive create, truncate, rename, unlink, the rmdir/unlink type assertions,
  statvfs, and the per-trace `newfs` recycle — as ordinary correctness.
- **SD2–SD5 as the defect.** An unexercised bug is indistinguishable from a
  fixed one, so each pin names what to assert instead once the defect is fixed,
  and fails loudly if the behaviour improves. The pins assert behavioural
  signatures (the op fails; the created object cannot be seen afterwards; fewer
  entries return than were created) rather than errno values, because the
  numbers this proxy produces are neither stable across platforms nor the ones
  POSIX defines here — itself part of SD5.

## The memory-safety fix

`chimera_smb_lookup_create_reply` documented that it returns "attrs but NO child
fh", but only left `va_fh` unwritten — and vfs requests come off a free list, so
`va_fh_len` still held the previous request's value. `chimera_vfs_lookup_at_complete`
reads that length to decide whether a handle came back, so garbage sailed past
the `> 0` test and became the size argument of the name cache's memcpy:

```
ERROR: AddressSanitizer: negative-size-param: (size=-1718575551)
  #2 chimera_vfs_name_cache_insert vfs_name_cache.h:272
  #3 chimera_vfs_lookup_at_complete vfs_proc_lookup_at.c:31
  #4 chimera_smb_lookup_close_reply smb_ops.c:491
```

Fixed on both sides: the smb reply zeroes the length, and the engine now gates
on the `ATTR_FH` set bit rather than on `va_fh_len` alone, so the next path-only
backend cannot reintroduce the overflow by simply not writing a field. The probe
executes that line 14 times per run.

## The MBT batch, parked

The harness also gains `smb_<module>` as a POSIX MBT backend, registered as
`chimera/posix/mbt/batch_smb_memfs` behind `-DCHIMERA_POSIX_MBT_SMB=ON` and
**off by default**: it fails 49 of 53 traces (2,916 divergences). That is not
instability — the SMB2 loopback cannot satisfy the POSIX model yet, and the five
blockers are recorded next to the test so turning the flag on measures progress
against a named list rather than against noise.

| | Defect |
|---|---|
| **SD1** | An SMB session is bound to the connection that authenticated it. A second client thread opens its own connection and replays the mount's session id; the server rejects it, mapped to ESTALE — so the first operation after the mount fails whenever it lands on another thread. The driver pins `core_threads=1`; SMB3 multichannel session binding is the real fix. |
| **SD2** | `readdir` returns exactly **one** entry per directory, whatever its size (verified at 1/2/3/5/10). The QUERY_DIRECTORY buffer is walked until the emit callback reports full, then the unconsumed remainder is freed; the resume query asks the server for the *next* batch, which is empty. Entries 2..N are lost. |
| **SD3** | `symlink` directly under the mount root reports **success and creates nothing**: `lstat` answers ENOENT, `readlink` answers ELOOP, and it is absent from readdir. |
| **SD4** | No POSIX owner or mode. `chmod`/`chown` are accepted and dropped, `getattr` reports the *calling* credential as owner and a synthesized mode, directory `nlink` is always 1 — so every EACCES/EPERM the model expects does not happen. |
| **SD5** | Only the mount-root handle is re-openable by fh, so every op routed through a resolved *parent* handle answers ESTALE: `link` and `utimens` at any depth, `symlink`/`mknod` below the first level. |

SD2 is the natural next PR: self-contained, clearly root-caused, and the largest
single obstacle to the batch running for real.

## Note on the corpus

No `smb_` trace corpus is generated. The model has no knob for "no owner, no
mode", so a pinned specs profile would assert a policy this backend does not
implement; the batch reuses the `nfs3_` corpus, whose measured profile differs
only in `copy_file_range`. `posix_replay.py` pins the probed profile for
`--check-profile` drift detection. Once SD2–SD5 are fixed, a dedicated
`posixSmb` profile in `ext/specs` is the right next step.

### DIFF
--- a/src/posix/tests/quint/CMakeLists.txt
+++ b/src/posix/tests/quint/CMakeLists.txt
@@ -126,6 +126,60 @@ set_tests_properties(chimera/posix/mbt/batch_nfs4_memfs PROPERTIES
     SKIP_RETURN_CODE 77
     TIMEOUT 600)
 
+# The same corpus through the SMB2 loopback: the posix client mounts the
+# in-process chimera SMB server through the smb proxy VFS module (src/vfs/smb).
+# The nfs3_ profile is the closest pinned match to what `posix_replay.py
+# --probe --backend smb_memfs` measures (they differ only in copy_file_range,
+# which the model exercises through an advertised-feature knob).
+#
+# OFF by default, and it is not a flake: the SMB2 loopback cannot satisfy the
+# POSIX model today, and the batch fails ~49 of 53 traces.  The blockers are
+# properties of src/vfs/smb, not of the harness:
+#
+#   SD1  a session is bound to the connection that authenticated it.  A second
+#        client thread opens its own connection and replays the mount's session
+#        id on it, and the server answers one of the session/connection-gone
+#        statuses, which the proxy maps to ESTALE -- so the first operation
+#        after the mount fails whenever it lands on another thread.  The driver
+#        pins core_threads=1 to work around it; SMB3 multichannel session
+#        binding is what would actually fix it.
+#   SD2  readdir returns exactly ONE entry per directory, whatever its size.
+#        The QUERY_DIRECTORY buffer is walked until the emit callback reports
+#        full, and the unconsumed remainder is freed; the resume query asks the
+#        server for the NEXT batch, which is empty, so entries 2..N are lost.
+#   SD3  symlink_at directly under the mount root reports SUCCESS without
+#        creating anything: lstat of the new name then answers ENOENT, readlink
+#        answers ELOOP, and it is absent from readdir.
+#   SD4  no POSIX owner or mode.  chmod/chown are accepted and dropped, getattr
+#        reports the CALLING credential as the owner and a synthesized mode
+#        (0755/0644), and directory nlink is always 1.  Every EACCES/EPERM the
+#        model expects therefore does not happen.
+#   SD5  only the mount-root handle is re-openable by fh (chimera_smb_open_fh),
+#        so every op the client routes through a resolved PARENT handle rather
+#        than through its path-op strategy answers ESTALE: link and utimens at
+#        any depth, and symlink/mknod below the first level.  ESTALE is also
+#        the wrong answer for an operation the backend simply does not
+#        implement -- mknod at the first level, which does reach the module,
+#        correctly answers ENOTSUP.
+#
+# Turn it on with -DCHIMERA_POSIX_MBT_SMB=ON to measure progress against that
+# list; it joins the quick tier only once it can pass.
+option(CHIMERA_POSIX_MBT_SMB
+       "Register the POSIX MBT batch over the SMB2 loopback (known failing)" OFF)
+if(CHIMERA_POSIX_MBT_SMB)
+    add_test(NAME chimera/posix/mbt/batch_smb_memfs
+        COMMAND $<TARGET_FILE:posix_mbt_replay>
+                --backend smb_memfs
+                --trace-dir ${SPECS_BUNDLE_DIR}/posix
+                --exclude-prefix step --exclude-prefix disk_
+                --exclude-prefix fuse_ --exclude-prefix nfs4_)
+    set_tests_properties(chimera/posix/mbt/batch_smb_memfs PROPERTIES
+        ENVIRONMENT "CHIMERA_MBT_ARTIFACTS=posix_batch_smb_memfs"
+        LABELS "quint;posix_mbt;memfs;smb_mbt"
+        SKIP_RETURN_CODE 77
+        TIMEOUT 600)
+endif()
+
 # The disk_ profile replayed against the host-passthrough backends.  The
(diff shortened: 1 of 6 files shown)


benjarvis:
Pushed `a30af4d4` in response to the coverage report showing `src/vfs/smb` at 0%.

To be clear about what that number was: it is **pre-existing, not introduced here** — PR #1609, on the same day, reports the same `src/vfs/smb` 0% (0/2,799 lines; this branch reads 0/2,800, the +1 being the line the fix adds). Nothing labelled `quint` executes the SMB *client* proxy: its only test, `smb_mount_test`, is gated behind `CHIMERA_LIMITS_TESTING`, and the MBT batch this PR adds is off by default.

The fair criticism stands regardless: the first commit landed a memory-safety fix on a line CI never runs. The patch-coverage row said so directly — `src/vfs/smb/smb_ops.c` 0% (0/1).

So this adds `chimera/posix/smb/loopback_probe`: a corpus-free, green, quick-tier probe over the same loopback. It asserts the operations the proxy does implement as ordinary correctness, and pins SD2–SD5 as the *defect* — each pin naming what to assert instead once the defect is fixed, and failing loudly if the behaviour improves, so a fix cannot land silently.

Measured over the probe alone:

| File | Lines |
|---|---|
| `smb_ops.c` | 0% → **83%** |
| `smb_mount.c` | 0% → **81%** |
| `smb_umount.c` | 0% → **81%** |
| `smb_io.c` | 0% → **80%** |
| `smb_ntlm.c` | 0% → **86%** |
| `smb.c` | 0% → **64%** |
| `smb_namespace.c` | 0% → **54%** |

`chimera_smb_lookup_create_reply` line 522 — the fix — now executes 14 times per run. Quick tier is 50/50 green.

https://claude.ai/code/session_01RvziibNjT2WwfThiPdmj9Z

benjarvis:
**Merge queue break — fixed, force-pushed as `a090d5b9`.** Apologies; this one was mine and it broke the batch on every platform.

It was a *build* failure, not a test failure:

```
FAILED: src/posix/tests/quint/CMakeFiles/smb_loopback_probe.dir/smb_loopback_probe.c.o
src/posix/tests/quint/posix_driver.c:78:35:
  error: 'proto_out' defined but not used [-Werror=unused-variable]
```

`proto_out` is a file-scope static in `posix_driver.c`, used only by the line-protocol `main()` that `POSIX_DRIVER_ENGINE_ONLY` compiles out. `posix_mbt_replay.c` includes the driver and *does* touch it, for its stdout/stderr split. My probe included the driver and did not — so gcc flagged it. Clang does not warn for this (unused static *variables* are `-Wunused-const-variable` there, const only), so it was invisible on macOS where I developed and tested.

The fix is to do what `posix_mbt_replay.c` already does: save the real stdout into `proto_out`, point fd 1 at stderr, then restore. That removes the warning by using the variable for its intended purpose, and as a bonus the probe's own output now lands on the real stdout instead of being folded into the log stream.

Two process notes, since this should not have reached the queue:

- **Why PR checks passed:** they build Release with clang on macOS and gcc on Linux — but the Linux pre-merge job builds only, and the break is in a test target that the pre-merge build set does not cover. It first compiled in the merge queue's full `Test <distro>` jobs.
- I've verified the fixed probe builds and the full quick tier is 50/50 green locally, but I want to be straight that my local toolchain (clang, macOS) cannot reproduce this class of gcc `-Werror` break. The fix is structural rather than empirically reproduced: the variable is now genuinely read and written, so "defined but not used" cannot apply under any compiler.

#1613 (stacked) has been rebased onto the corrected commit and force-pushed.

https://claude.ai/code/session_01RvziibNjT2WwfThiPdmj9Z

PR #1613, opened 2026-09-01
smb: pin the client's dialect with vers=, and cover the whole matrix

benjarvis:
## The gap

The SMB client offers every dialect it can carry on every NEGOTIATE, and the
server picks the highest in common. Against our own server that is always
**3.1.1**, verified on the wire:

```
SMB mount established: //127.0.0.1/share (dialect 0x0311, session ..., tree 1)
```

So 2.1, 3.0 and 3.0.2 were unreachable in practice — different signing-key
derivation, different signing algorithms, and a NEGOTIATE with no contexts at
all, none of it ever executed no matter how many tests ran.

To answer the question directly: the client advertises **4 of the 5** dialects
(`CHIMERA_SMB_CLIENT_DIALECTS` = 2.1, 3.0, 3.0.2, 3.1.1). It does **not** carry
2.0.2 — `chimera_smb_client_negotiate_reply` refuses it — though the server
speaks all five. Until this PR, only one of those four had ever been exercised.

## What this adds

`vers=` as a client mount option. A pinned mount offers exactly **one** dialect,
so the server can only select that one or fail to find one in common — a mount
that comes up is *proof* the connection is speaking the dialect asked for, with
no readback to trust.

Two details the pinned path needs:

- negotiate contexts (preauth integrity, signing capabilities) exist only from
  3.1.1; sending them while offering an older dialect is a protocol error, so
  they are emitted only when 3.1.1 is on offer.
- an unrecognized `vers=` is EINVAL, not a silent fallback. A quietly-ignored
  option would come up on 3.1.1 and look like it had tested a dialect it never
  negotiated.

`vers=` accepts the dotted names an administrator would write (`2.1`, `3.0`,
`3.0.2`, `3.1.1`, plus the `3.02`/`3.11` spellings mount.cifs takes) or a raw
`0xNNNN`.

## Results

The loopback probe now runs once per dialect. **All four pass the full probe**,
each verifiably negotiating the dialect named:

| vers= | negotiated | probe |
|---|---|---|
| `2.1` | 0x0210 | pass |
| `3.0` | 0x0300 | pass |
| `3.0.2` | 0x0302 | pass |
| `3.1.1` | 0x0311 | pass |
| `2.0.2` | — | mount refused (expected; client does not carry it) |

`loopback_dialect_refused_2.0.2` is the companion negative and is what proves
`vers=` is honoured at all: an ignored option would negotiate 3.1.1, mount
happily, and fail that test.

Coverage of `src/vfs/smb` from the probe alone, over #1610's single-dialect
baseline: `smb.c` 64% → **71%**, `smb_mount.c` 81% → **82%** — the signing and
no-context negotiate paths.

## A crash the negative test walked into

Building a test whose *result* is "the mount was refused" exposed a shutdown
race in the harness. A failed mount left both the in-process server and the
client holding live evpl thread pools; returning from `posix_env_setup` without
stopping them races process teardown, and those threads keep entering
`evpl_run` while the global evpl state is destroyed underneath:

```
SEGV on unknown address 0x0000000000c8
  #0 evpl_now_ticks evpl_shared.h:70
  #1 evpl_run evpl.c:797
  #2 evpl_thread_function thread.c:109
```

Measured at **2 crashes in 10 runs**. `posix_env_setup_unwind` stops the server
first (its threads are the ones that crash) and then the client: **0 in 20**
after. Worth noting the obvious wrong fix — shutting the *client* down alone
made it deterministic at 20/20, because that is what pulls the shared evpl
state out from under the still-running server threads.

Quick tier: 55/55 green.

### DIFF
--- a/src/posix/tests/quint/CMakeLists.txt
+++ b/src/posix/tests/quint/CMakeLists.txt
@@ -49,6 +49,36 @@ set_tests_properties(chimera/posix/smb/loopback_probe PROPERTIES
     LABELS "quint;posix_mbt;memfs;smb_mbt"
     TIMEOUT 300)
 
+# The same probe once per SMB dialect the client can carry.  Without a pinned
+# vers= the server selects the highest dialect in common, which against our own
+# server is always 3.1.1 -- so 2.1, 3.0 and 3.0.2 were dead code in practice no
+# matter how many tests ran.  Pinning is the only way to reach them.
+#
+# A pinned mount offers exactly ONE dialect, so a mount that comes up is proof
+# the connection is speaking that dialect: the server can only select the one
+# offered or fail to find one in common.  There is no readback to trust.
+foreach(vers 2.1 3.0 3.0.2 3.1.1)
+    add_test(NAME chimera/posix/smb/loopback_probe_${vers}
+        COMMAND $<TARGET_FILE:smb_loopback_probe> ${vers})
+    set_tests_properties(chimera/posix/smb/loopback_probe_${vers} PROPERTIES
+        ENVIRONMENT "CHIMERA_MBT_ARTIFACTS=smb_loopback_probe_${vers}"
+        LABELS "quint;posix_mbt;memfs;smb_mbt"
+        TIMEOUT 300)
+endforeach()
+
+# 2.0.2 is the dialect the client does NOT carry: it is absent from
+# CHIMERA_SMB_CLIENT_DIALECTS, and chimera_smb_client_negotiate_reply refuses it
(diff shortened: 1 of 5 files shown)


benjarvis:
Rebased onto main now that #1610 has merged. The diff is down to this PR's own two commits — `smb: let a client mount pin its SMB dialect with vers=` and `tests/quint: run the SMB loopback probe against every dialect` — and the stacking caveat is gone from the description. Quick tier 57/57 green on the new base.

PR #1664, opened 2026-09-08
vfs/smb: store native Windows SIDs across the VFS boundary

arenaud16:
A principal was a uid/gid, so a SID the identity layer could not map had no uid
and was dropped on the way in. A Windows client that set a DACL naming a domain
user this server has never heard of read the descriptor back with that ACE
missing -- including DENY aces, whose disappearance grants access rather than
denying it. The same was true of the owner and group fields, which were always
re-derived from uid/gid on the way out.

This series carries the native SID alongside the numeric identity, end to end,
so a Windows ACL survives a round trip through a POSIX-backed share.

Access evaluation is deliberately untouched: it still runs purely on the
numeric identity. A SID that resolves to nobody matches no caller, which is the
same thing NTFS does with a departed user's ACE.

### The value type and the SDK

`struct chimera_sid` (`src/vfs/sdk/vfs_sid.h`, `vfs_sid.c`) holds a SID in its
MS-DTYP binary form with the 15 sub-authority cap, and owns the binary/string
codec that `smb_proc_security.c` previously kept as file-local statics. The
string parser is strict -- digits only, no trailing text -- so a malformed wire
SID is rejected rather than overrunning a buffer.

`struct chimera_principal` gains an inline SID. It is populated only on USER and
GROUP principals that resolved through the identity layer, and on the new
`CHIMERA_PRINCIPAL_SID` for a SID that resolved to nothing at all; SPECIAL
principals never carry one, since `OWNER@` and friends re-derive on emit.

The principal is a published SDK type that out-of-tree modules consume, so
`CHIMERA_VFS_SDK_VERSION` moves to 2 and a module built against 1 must be
rebuilt.

On-disk ACL encoding gains a v2 that appends a length-prefixed SID to each ACE.
The serializer picks v2 only when some ACE actually carries a SID and otherwise
still writes v1, so a volume stays readable by an older binary until a real
domain SID is stored; the deserializer accepts both.
`chimera_acl_serialized_size` becomes a walk, since the encoding is now
variable-length.

### Attributes and the chown gate

`CHIMERA_VFS_ATTR_OWNER_SID` / `GROUP_SID` and the pointer-valued
`va_owner_sid` / `va_group_sid` are the SID companions to `va_uid` / `va_gid`,
with `va_acl`'s callback-lifetime contract and deliberately outside `MASK_STAT`
and `MASK_CACHEABLE`.

Two consequences are worth calling out. The setattr gate folds the SIDs into
its "is this really a chown" test the same way it already treats a restated
uid, so a SET_INFO security descriptor that restates the owner SID the object
already carries alongside a new DACL keeps requiring only `WRITE_ACL`, which
the owner holds implicitly. And the attr cache copies the whole attrs struct by
value, while the ACL and SID pointers in that copy reference per-thread backend
scratch that is dead once the producing callback returns -- so the cache now
clears those bits and NULLs those pointers on insert.

### Backends

memfs, cairn and diskfs all keep the owner/group SIDs in a record *separate*
from the ACL, so "no ACL record" keeps meaning "mode-derived DACL" and a chown
never has to touch the DACL record. cairn adds `CAIRN_KEY_SID`, diskfs adds
`DISKFS_REC_SID` with the same mirror discipline as its ACL record, and memfs
stores them on the inode.

All three keep the SIDs coherent with uid/gid: an explicit `OWNER_SID` /
`GROUP_SID` set stores or clears, while a bare chown drops the stale one that
would no longer describe the new owner.

Because a v2 ACE is variable-length, whether an ACL fits one diskfs record is
now decided by `chimera_acl_serialized_size` rather than an ACE count. diskfs
has always persisted the canonical ACL but never advertised
`CHIMERA_VFS_CAP_ACL_NATIVE`; it does now, like the other two.

### SMB

SET security keeps what it used to discard. An owner or group SID that resolves
travels as the `va_owner_sid` / `va_group_sid` companion; an ACE SID that
resolves keeps the SID on its USER/GROUP principal; and an ACE SID still
unmappable on the final resolve pass is stored as an opaque
`CHIMERA_PRINCIPAL_SID` rather than dropped. The create-context SD parse gets
the vfs too, so the session's own cached SID resolves at create time while
uncached ones are preserved verbatim.

QUERY security prefers what was stored: the owner and group fields use the
native companions when present, an ACE whose principal carries a SID is emitted
from those bytes, and an `OWNER@` / `GROUP@` entry substitutes the stored
owner/group SID so it agrees with the descriptor's own owner and group fields.
Only SID-less principals still walk the identity cache and fall back to the
algorithmic `modefromsid` form. The uncached-principal count that decides
whether to park for resolution skips anything already carrying a SID, so a
descriptor built from stored SIDs never waits on the resolver.

`sec_buf` grows 2048 -> 4096, since a 64-ACE descriptor carrying real domain
SIDs no longer fits 2 KiB.

### Coverage

`vfs/sid_test`, `vfs/acl_sid_*` across memfs/cairn/diskfs and `smb/sd_sid_test`
are all extended-tier, and the merge gate runs the quick tier -- so on their own
they would not catch a regression here until the next Extended run. The SMB2
trace corpus cannot cover it either: the quint model has no security-descriptor
surface at all, and `probe_security()` stopped at the 20-byte header without
ever decoding a SID or walking an ACE. Its own comment said the translation
layer was "otherwise dark".

The last commit closes that, in the probe that already owns this surface and
already runs in the quick tier. Following the house discipline of pinning
behaviour with a probe before a model bakes an expectation, it covers the case
that actually changed: a descriptor the server did *not* author. It SETs a DACL
carrying an unmappable domain user, a well-known SID and a DENY ace for a
second unmappable user, then requires all three back verbatim with type, access
mask and order intact. Built without the series the same probe fails, dropping
two of the three aces, the DENY among them.

It also pins the two boundaries: a file with nothing stored still reports the
algorithmic `S-1-5-88` owner and group, and an owner SID no identity authority
can resolve is deliberately *not* adopted -- silently taking one would detach
the file's owner from every POSIX check.

`smb2_sd_parse` / `smb2_sd_build` / `smb2_sid_{to_str,str_to_bin}` go in
`smb2_mbt_common.h` rather than in the probe, since nothing in the quint
harness could decode a SID or an ACE and a replayer handler for a modeled
SECURITY op will need exactly this.

### Notes for review

Deliberately out of scope, and worth saying so explicitly:

- Owner/group SID companions are not seeded at create.
- An unmappable OWNER/GROUP SID on SET is ignored rather than refused with
  `STATUS_INVALID_OWNER`.
- A stored `PRINCIPAL_SID` never upgrades if the SID later becomes resolvable.
- NFSv4 who-strings are unchanged; this is SMB-only on the protocol side.
- Modeling the security-descriptor surface in the shared quint spec is a
  separate piece of work. It needs a `READ_CONTROL` / `WRITE_DAC` /
  `WRITE_OWNER` axis on `DesiredAccess` that the model does not have today,
  without which every modeled `CSetSecurity` would return
  `STATUS_ACCESS_DENIED`.

One divergence noticed while working here but not changed: chimera does not
check `READ_CONTROL` on QUERY SECURITY, where MS-SMB2 gates it.

### DIFF
--- a/docs/smb-domain-auth.md
+++ b/docs/smb-domain-auth.md
@@ -50,6 +50,30 @@ The POSIX uid and gid that chimera stamps on files created over SMB come from
 winbind's identity mapping, so the idmap backend you choose decides on-disk
 ownership.
 
+### Native SIDs in stored security descriptors
+
+Access decisions are always made from the numeric uid/gid, but the security
+descriptor a client sets is not reduced to numbers and thrown away.  When a
+SID in an incoming descriptor resolves through winbind, chimera keeps the
+native SID alongside the resolved id: on the owner and group as SID
+companions to the uid/gid, and on each DACL entry's principal.  A SID that
+does not resolve at all is kept verbatim as an opaque principal that grants
+nobody anything -- the same way NTFS keeps an entry for a departed domain
+user -- instead of being dropped from the DACL.  The native backends (memfs,
+cairn, diskfs) persist those SIDs, and a QUERY returns them byte for byte
+without consulting winbind again, so a stored descriptor survives a winbind
+outage and can be read by any other consumer of the on-disk data.
+
+Only when no native SID is known does chimera fall back to the identity
+cache and then to the algorithmic `S-1-5-88-1-<uid>` / `S-1-5-88-2-<gid>`
+scheme.  A plain chown or chgrp (over NFS, say) drops the stored owner or
+group SID, since it no longer describes the new numeric owner.
+
+The on-disk ACL encoding moved to version 2 to carry the per-entry SID;
+version 1 data written by earlier releases still reads, and an ACL with no
+native SIDs is still written as version 1.  Out-of-tree VFS modules must be
+rebuilt against SDK version 2.
+
 ### Configuration keys
 
 The chimera side of both procedures is the `server.smb_auth` object:
--- a/docs/vfs-module-sdk.md
+++ b/docs/vfs-module-sdk.md
@@ -51,8 +51,9 @@ xxhash comes from the system (`libxxhash-dev` / `xxhash`).
 | `vfs_log.h` | `chimera_vfs_debug/info/error/fatal/abort` macros, self-contained |
 | `vfs_fh.h` | **The file-handle routing contract** (see below) and the encoders that satisfy it |
 | `vfs_varint.h` | Varint primitives the inum-style file-handle encoders are built on |
-| `vfs_acl.h` | `struct chimera_acl` / `struct chimera_ace` — the canonical ACL carried by `chimera_vfs_attrs.va_acl` |
-| `vfs_acl_serialize.h` | Stable, versioned encoding of a `chimera_acl` for a backend's own persistence |
+| `vfs_sid.h` | `struct chimera_sid`, the native Windows SID value type (MS-DTYP binary form) carried on ACL principals and on the owner/group attrs, plus its binary/string codec |
+| `vfs_acl.h` | `struct chimera_acl` / `struct chimera_ace` — the canonical ACL carried by `chimera_vfs_attrs.va_acl`; each principal pairs a numeric uid/gid with an optional native SID |
+| `vfs_acl_serialize.h` | Stable, versioned encoding of a `chimera_acl` for a backend's own persistence (v2 adds the per-ACE native SID; v1 blobs still decode) |
 | `vfs_access.h` | Access-mask evaluation, so `ACCESS` answers agree across backends |
 | `vfs_xattr_name.h` | The protocol-exported `user.` xattr keyspace NFS and SMB both normalize into |
 | `vfs_tcp_flavor.h` | `enum chimera_tcp_flavor`, the value `chimera_vfs_request_tcp_flavor` returns |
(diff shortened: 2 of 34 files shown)


arenaud16:
Closing this in favour of a smaller stack. At ~5k lines in one PR it never drew
a review, and it is easier to justify a piece at a time.

The same work, rebased onto current main and cut three ways so each piece
stands on its own:

1. **#1675** `smb: compare security-descriptor offsets before adding to them` --
   open now. A pre-existing wrapping-offset out-of-bounds read in the SD
   decoders, reachable by any authenticated client and independent of
   everything else here.
2. `vfs: store native Windows SIDs across the VFS boundary` -- the SID value
   type, SDK version 2, v2 ACL serialization, the owner/group SID attrs and the
   chown gate, memfs/cairn/diskfs, and the unit tests. No SID reaches a client
   from this half alone.
3. `smb: carry native Windows SIDs through security descriptors` -- SET and
   QUERY, the validate-then-commit decoder and its refusal contract, the docs,
   the SD codec tests and the quint probe, which is the first quick-tier
   coverage this surface has had.

2 and 3 will go up as their own PRs. Nothing is dropped: between them the three
branches carry everything that was on this one, plus the fixes from a review
round -- NFSv4-set ACEs no longer carrying an uninitialised SID, opaque SID
principals no longer marshalling to NFSv4 clients as uid 0, SET refusing an
oversized or malformed descriptor as a whole instead of applying a subset, and
every diskfs setattr path converging on one record tail.

PR #1679, opened 2026-09-11
Revert "smb: apply a break the dying channel can no longer deliver, when an opener is waiting"

benjarvis:
Reverts a9a50dd9 (#1673). **My change, my regression.**

It broke `smb2.durable-open.lease-disconnect-race` — the exact suite it was supposed to keep green, with the exact assertion two earlier attempts died on:

```
source4/torture/smb2/durable_open.c:2262:
  io2.out.lease_response.lease_state was 3 (0x3), expected 7 (0x7)
```

## Evidence

`smbtorture_smb2_durable_open_{lease,oplock}_disconnect_race_memfs` fail in the `rest` shard on **all three** of amd64 Release, arm64 Debug and arm64 Release, in **both** extended runs since it merged:

| run | head | contains #1673 | rest-shard failures |
|---|---|---|---|
| 34504834422 | `98e7a0a3` | no | **0** |
| 34531970287 | `49daa5c7` | yes | **3** |
| 34566645704 | `49daa5c7` | yes | **3** |

Reproduced locally and A/B'd behind a kill-switch on the `conn_free` break-resolution block:

| arm | failures |
|---|---|
| with the change | **10 / 36** |
| with it disabled | **0 / 36** |

After this revert: **0/36** concurrent rounds, and `smbtorture_smb2_{durable,lease,oplock,dirlease}_*_memfs` back to **143/143**.

## Why my original verification missed it

I ran the 143 smbtorture tests serially, twice, and they passed — as they still do: 20/20 serial rounds pass even with the bug in. The failure only appears under the parallelism the `rest` shard actually uses (`ctest -j $(nproc)`). My WPTS gate was correctly run concurrently; the smbtorture gate was not. The handoff's warning was to run both gates, and I did — but not both *the way CI runs them*.

## What this costs

The WPTS `ReplayCreateDurableHandleV2Persistent` intermittent comes back — roughly one case in fifty, in one shard, some runs. That is strictly cheaper than three jobs failing deterministically every night.

## What survives

The park/resume/deadline logging (02661f28, #1671) is untouched — it changes no behaviour and is what produced the root cause. The diagnosis stands: a break notification pinned to `create_conn` is orphaned when that channel dies while the session lives on.

What does not stand is my reconciliation. Gating on "is an opener parked on this break" narrows the window but does not resolve the conflict: `smb2.durable-open.lease-disconnect-race` still requires the break be abandoned in cases my guard treats as waiters. The next attempt needs the concurrent smbtorture loop as a gate from the start — it reproduces in ~10/36 and is cheap to run.


benjarvis:
Superseded by #1681, which fixes the regression in place rather than reverting — strictly better than a revert on the WPTS rate (1/65 failing shards vs 4/50).

PR #1680, opened 2026-09-11
smb: re-land the undeliverable-break fix, gated on the holder still being reachable

benjarvis:
**Stacked on #1679** (the revert). Merge that first; this branch contains it, so the net diff against today's `main` is the corrected version of #1673.

## What was wrong with the first attempt

The root cause is unchanged and still stands: a break notification is pinned to `open_file->create_conn`, the session outlives the channel under multichannel, and `conn_free` clears the stale pointer while leaving an in-flight break unresolved. The claim sits `BREAKING` for the 30 s deadline; the WPTS adapter gives up at 20 s.

What #1673 got wrong is **which** breaks to resolve. It gated on *"is an opener parked on this break"*, reasoning that nothing is parked in `smb2.durable-open.lease-disconnect-race` because the holder disconnects before any opener exists. Under the parallelism that suite actually runs at, that's false often enough to matter — the opener's CREATE and the holder's disconnect interleave, the guard fires, and a break is applied to a holder entitled to keep its lease:

```
durable_open.c:2262: io2.out.lease_response.lease_state was 3 (0x3), expected 7 (0x7)
```

**10 failures in 36 concurrent runs.** My original verification ran that suite serially, where it passes 20/20 even with the bug — the WPTS gate was run concurrently, the smbtorture gate was not, and that asymmetry is the entire miss.

## The corrected discriminator

Not *who is waiting* — whether the holder is still **reachable**:

| | meaning | action |
|---|---|---|
| session has another bound channel | holder is alive, merely unreachable on *this* transport; it never got the break and can't be told on the dead channel | resolve the break |
| session lost its **last** channel | holder has genuinely gone away; its durable handle is preserved for reconnect and must return with the full lease it never gave up | leave it — the deadline still bounds it |

That is precisely the difference between the two suites: WPTS exercises multichannel, the durable-open race tests a plain single-channel disconnect.

`num_channels` already carries this, and `conn_free` reads it *before* releasing the session handles that decrement it, so the dying channel is still counted — a sibling exists iff `num_channels > 1`. One condition, no new state.

## Validation — both gates, each run the way CI runs it

| gate | result |
|---|---|
| `durable-open.{lease,oplock}-disconnect-race`, 6 concurrent × 6 rounds | **0/36** (reverted version: 10/36) |
| `smbtorture_smb2_{durable,lease,oplock,dirlease}_*_memfs` | **143/143** |
| WPTS `multichannel_persistent_encryption`, 5 shards concurrently | **15 rounds / 75 shard runs clean** |
| ↳ positive control (this fixup disabled) | **reproduces in round 1** |
| local quint/MBT smb suites | **11/11** |

The positive control matters: an earlier reading of mine took a clean WPTS loop as proof the fix worked, when the wrapper only dumps the daemon log on failure — so "no park lines" meant nothing. The control establishes the gate is live before the clean arm is worth anything.

Clean on macOS Debug/ASan and Linux Release; uncrustify 0.78.1 and copyright checks pass.


benjarvis:
Superseded by #1681 — same fix, rebased onto current main as a single minimal commit rather than stacked on a revert.

PR #1689, opened 2026-09-14
vfs/diskfs: recall bump reservations instead of reporting a false ENOSPC

benjarvis:
Targets the `nfstest_alloc_diskfs_{aio,io_uring}_nfs4.2` cells, which have failed **5 runs out of 5** on the same SHA — the only deterministic product failure left in Extended.

## The bug

A bump reservation claims a region of an AG that the owning thread hands out locally. The claim **does not remove the region from the AG free tree** — it only stops another thread re-handing it. So a claim's unused tail reads as *free* to statfs while being reachable only by its owner. An idle worker parks `SM_RESERVATION_CHUNK` indefinitely, and a request for one block fails with megabytes still reported free.

At the geometry the kvm shards use (2 × 128 MiB behind a 64 MiB intent log, 2 workers) that is **8 MiB stranded**, and a fallocate of all free space stopped **58 MiB short**.

## Recall

When a grab finds no claim-free window anywhere, `space_map_reserve_chunk` makes a second pass that hands back the unallocated part of each claim, retrying **per AG** rather than recalling everything first — so a thread still working loses no more than the request needs. Only the path that was about to return ENOSPC pays for it.

To make a claim's spoken-for extent visible to other threads, the bump cursor moves from `struct sm_reservation` into `struct sm_claim` alongside the limit, both atomic. Each has a single writer — the owner advances the cursor, a recaller (under `ag->lock`) lowers the limit — which keeps the owner's bump **lock-free**: no CAS, just a re-read of the limit after publishing the new cursor, backing out if a recall landed in between. The recaller concedes symmetrically if the owner bumped while it was lowering. Both sides carry the interleaving argument in comments.

## Cap

A grant is now bounded by a share of the AG's remaining free space (`SM_RESERVE_AG_SHIFT`) as well as by `SM_RESERVATION_CHUNK`, so reservation size follows the AG down — 4 MiB, then 1 MiB, then 128 KiB — converging on the exact ask. Never below the caller's `want`.

`space_map_reserve()` already applies this to the *other* reservation mechanism (the thread cache) as a binary *"don't speculate when free < 2×want"*; this is the graduated form for bump claims.

**Cap alone does not fix it** — the stranded claims are grabbed while the AG is still full, so the cap never binds on them. Measured: still 58 MiB short with the cap alone. Recall is the corrective mechanism; the cap governs convergence during the fill.

## A trap worth flagging for review

Recall alone still failed, subtly: a claim trimmed to its cursor can have **length zero**, and a zero-length claim overlaps nothing (so the overlap scan skips it) while still sitting in `sm_ag_next_claim_base_locked`'s barrier scan. That capped grants at the empty claim's base and returned **zero-length reservations that callers read as success** — surfacing as `post-ensure bump FAILED ... len=0`. Empty claims are now skipped as barriers, and `sm_ag_try_claim_locked` refuses to return a grant below `want`.

## statfs

With the allocator able to reach what it advertises, the internal reserve now comes off `space_free` as well as `space_avail`.

The ext4/XFS framing behind #1685 only goes so far: their reserved blocks are a **privilege** reserve that root may spend, so counting them in `f_bfree` is truthful. Ours is metadata slop no caller can turn into file data, so reporting it in `f_bfree` overstates what is obtainable — and a root caller that believes `f_bfree`, which is exactly what `nfstest_alloc`'s `get_freebytes()` does, asks for space that cannot exist. That is why #1685 did not move the needle.

## Results

| | before | after |
|---|---|---|
| allocator reaches (of 184197120 free) | 125549497 | **184167608** |
| short by | ~58 MiB | **~25 KB** (extent-record metadata) |

The in-process alloc06 analogue now allocates everything statfs reports, so its `KNOWN GAP` branch becomes a **hard assert**.

- quick tier **113/113**
- `chimera/vfs/diskfs/*` **20/20** incl. MBT batch + smoke
- full-stack `chimera/posix/diskfs_enospc_diskfs` passes
- uncrustify 0.78.1 (CI's pin) PASS on all four files; copyright-year clean

The kvm `nfstest_alloc` cells need Linux + KVM, so the Extended run is their real gate.


benjarvis:
**Merge queue caught a real bug in this PR — fixed in `a33a03ee`.**

```
double-free or overlap at offset=42168320 length=4096 (prev=41889792+25219072)
space_map.c:392  fatal
```

`chimera/vfs/diskfs/smoke_test` aborted on ubuntu22 Debug/Release and ubuntu26 Release. Two ALLOC deltas for one block: the first apply removes it from the free tree, and the second file to release it frees a range that is already free.

### Cause: the synchronisation, not the idea

Cursor and limit lived in **separate words** — the owner stored the cursor then loaded the limit, a recaller stored the limit then loaded the cursor. That is Dekker's algorithm, correct only under sequential consistency. Written with release/acquire it permits precisely the store-load reordering x86 allows, so both sides could read stale, both conclude they had won, and the owner kept an allocation the recaller had already handed out.

Every path in that handshake was individually defensible, which is how it passed review and 113/113 of the quick tier locally — the Linux jobs use a different device backend and enough concurrency to hit it. I could not reproduce it on arm64 even at 24-way.

### Fix: one word instead of two

Both halves now live in a single atomic word as AG-relative 32-bit offsets (an AG is `SM_AG_SIZE` = 2 GiB, so each fits). Advancing the cursor and lowering the limit become compare-and-swaps on the **same word** — exactly one wins, the loser re-reads. The owner's back-out path disappears (a failed CAS just retries), and with it the need to reason about memory order at all.

I deliberately chose this over simply promoting the four operations to `SEQ_CST`. That would also have been correct, but it would have left the design resting on an ordering argument that had already fooled me once. A tighter invariant beats a longer argument.

### Verification

- quick tier **113/113** — taking ctest's own exit status this time, not a pipeline's (`tail` was masking it)
- diskfs suite 20/20; enospc boundary assert still holds
- 24 concurrent `smoke_test` runs clean
- uncrustify 0.78.1 PASS

benjarvis:
**Second merge-queue failure, and my first diagnosis was wrong.** Fixed in `153e101d`.

Same abort, same test:

```
double-free or overlap at offset=42168320 length=4096 (prev=41889792+25219072)
chimera/vfs/diskfs/smoke_test (Subprocess aborted)
```

### Why the memory-ordering reading was wrong

Two things ruled it out once the second run came in:

- the offsets repeat **byte for byte** across runs and across distros — a reordering race would corrupt somewhere different each time
- the **failing jobs move** while the corrupted block does not: run 1 was ubuntu22 Debug/Release, ubuntu26 Release, rocky10 Release; run 2 was devcontainer Release and rocky9 Debug, and only 2 of 24 jobs

That is a deterministic workload tripping a racy *trigger*, not a torn read. I'm keeping the single-word claim state from `a33a03ee` regardless — it is a tighter invariant than the two-word handshake it replaced, and it is what turns "recalled out from under us" into a clean bump failure the new loop can retry — but it was not the cause.

### The actual cause

`space_map_reservation_alloc` grabbed once and then assumed:

```c
if (space_map_bump_alloc(...) != 0) {
    return -1;      /* a freshly-grabbed chunk >= need must satisfy */
}
```

That held while a reservation was private to its thread. **Recall breaks it.** A newly granted claim has its cursor at its base, so every byte of it reads as unspoken-for — and a recall on another thread can take the *whole claim* back in the window between the grab and this thread's first bump.

The result is a **spurious ENOSPC**, raised mid-operation with the pool not full. `diskfs_enospc_test`'s own comment says where that ends:

> the file then references space the allocator believes is free, which the next allocation can hand to somebody else and **which double-frees the pool when the file is deleted**

Which is the abort, exactly.

### Fix

Loop instead of grabbing once: only a grab that genuinely finds no space — `ensure()` failing *after its own recall pass* — is ENOSPC. Losing the race is retried, bounded by `SM_RESERVATION_GRAB_RETRIES` so a recall storm cannot spin forever.

### Verification, and its limits

- quick tier **113/113** (ctest's own exit status)
- `smoke_test` **24/24** clean in a container configured with the merge queue's own flags (`-DIO_URING_ENABLED=OFF -DKVM_TESTING=OFF`, Debug), plus a full 131-test suite run at `-j nproc`
- `enospc` boundary assert still holds

Being straight about this: **I never reproduced the failure locally — 65 samples, all clean.** So these numbers show no regression, they do not prove the fix. CI is the gate. If it fails a third time I'd rather split the cap and the recall into separate PRs and bisect than keep guessing.

benjarvis:
**Bisected. The recall was never the cause — the cap was.** Ready to queue again (`59515529`).

Rather than attempt a fourth fix, I split this PR's two halves and ran the **full 24-job CI matrix** on each via `workflow_dispatch`, so neither run touched the merge queue:

| branch | contents | result |
|---|---|---|
| `bisect-cap-only` | 17 lines: `ag_cap = ag->free_bytes >> 3`. **No recall.** | [**FAILURE**](https://github.com/chimera-nas/chimera/actions/runs/34970868823) — same `double-free or overlap at offset=42168320`, same offsets |
| `bisect-recall-only` | recall, packed bump word, grab retry. **No cap.** | [**SUCCESS — 24/24**](https://github.com/chimera-nas/chimera/actions/runs/34970872307) |

So all three of my earlier diagnoses were aimed at the wrong component. For the record, they were: a Dekker memory-ordering race (wrong — offsets repeated byte-for-byte, which reordering would not do), and a spurious mid-operation ENOSPC from the grab retry (also wrong — recall alone is green).

## What changed here

The cap is dropped. It buys nothing on its own — recall is what makes the pool reachable to its last extents, and the cap measured no improvement without it — and it reproduces a `space_map` double-free that has nothing to do with reservations. The header records the measurement so the idea is not retried blind.

Filed separately as **#1694**: the cap changes no allocator invariant, only grant *size*, so reproducing a double-free from it means it is surfacing a **latent** bug in the space map. `bisect-cap-only` is the reproducer.

The other two commits stay, and are still wanted:
- **`a33a03ee`** packed bump word — a tighter invariant than the two-word handshake it replaced
- **`153e101d`** grab retry — genuinely required now, since a claim *can* be recalled out from under its owner between the grab and its first bump

## Verification

- `bisect-recall-only`, which is this content modulo the cap, is **green on all 24 CI jobs** — the first time this change has had a clean full matrix
- `enospc` passes with the boundary allocation succeeding, which is the behaviour this PR exists for
- `smoke_test` passes; uncrustify 0.78.1 PASS on both files

PR #1697, opened 2026-09-16
nfs4: fix two reboot-grace defects on the persistent-KV recovery path

arenaud16:
Two independent grace-window defects on the persistent-KV recovery path. Both
are unreachable on the default `memkv` backend, where `persistence_disabled`
short-circuits the whole mechanism, and become live as soon as `server.kv_module`
names a durable module.

### 1. A hydrated session's client is never reclaim-complete

A server restarted with a persistent KV module rebuilds a 4.1 session lazily,
the first time a client presents its old sessionid to SEQUENCE. The client is
held off with `NFS4ERR_DELAY`, the session and its client are reconstructed from
the store, and the retry succeeds. The point of that machinery is that the
client never learns the server restarted: it sees no `NFS4ERR_BADSESSION`, so it
runs no EXCHANGE_ID and no CREATE_SESSION.

It therefore never sends RECLAIM_COMPLETE, because nothing in the protocol asks
it to -- `SEQ4_STATUS_RESTART_RECLAIM_NEEDED` is never emitted anywhere in the
tree, so there is no channel by which it could be told to. But
`nfs4_drc_ensure_client()` calloc'd the restored record and left
`nfs4_client_reclaim_complete` clear, and the per-client gate in
`chimera_nfs4_open()` refuses a non-reclaim OPEN from a 4.1 client that has not
sent one. That gate carries no deadline -- unlike the server-wide reboot window,
which expires on the lease tick -- so every OPEN, and with it every read and
write, was refused `NFS4ERR_GRACE` for the remaining life of the mount. Metadata
operations are not gated, so the mount went on looking healthy while no file on
it could be opened.

RFC 8881 section 18.51.3 attaches the RECLAIM_COMPLETE obligation to
establishing a new client ID. A client resuming a reconstructed session
established nothing; it is the same confirmed client it was before the restart.
Restore the mark along with the rest of the record.

The server-wide grace window still applies and still expires normally, so a
client that resumes this way waits it out before its first OPEN lands, which is
correct: the in-memory open state is gone and the other clients get their chance
to reclaim first.

Two consequences of the mark are accepted deliberately and recorded in the
comment at the mark. The per-client gate now also refuses a CLAIM_PREVIOUS from
such a client with `NFS4ERR_NO_GRACE`. That forfeits nothing today: no open or
lock state is persisted, so a reclaim could rebuild nothing a post-grace
CLAIM_NULL cannot, and a client that re-establishes with the same owner and
verifier is answered `CONFIRMED_R` on the same clientid, which tells it no state
was lost, so it does not reclaim anyway. And the hydrated client's own
`to_reclaim` record can no longer be retired by RECLAIM_COMPLETE, because the
handler answers `NFS4ERR_COMPLETE_ALREADY` before it reaches the recovery
bookkeeping; only the grace deadline or DESTROY_CLIENTID retires it, so the
window cannot end early on the all-reclaimed exit while such a record is
outstanding. A hydrated client never sent a RECLAIM_COMPLETE, so no outcome
changes there either.

### 2. The grace window can close before the cold-start load has counted

`nfs_recovery_kickoff()` forces the window open and only then issues the epoch
read and the record scan that populate `to_reclaim`, so `pending_reclaim` is
still zero for as long as those two KV round trips take.
`nfs_recovery_sweep_once()`, which runs off the 1 Hz lease tick, reads a zero
count as "every reloaded client has finished reclaiming" and closes the window.

A tick landing in that gap therefore collapses the entire grace period to
whatever was left of one second, and the CLAIM_PREVIOUS reclaims the window
exists to accept are answered `NFS4ERR_NO_GRACE` instead -- the client loses
locks it was entitled to get back, silently. Two KV round trips against a cold
backing store span a 1 Hz tick without difficulty.
`nfs_recovery_finalize_load()` never re-arms the window either, so the damage
outlasts the load that caused it.

Only honour the zero-count exit once the load reports READY, which is the point
at which the count means what the sweep assumes. The deadline exit is unchanged
and still bounds the window either way.

`nfs_recovery_reclaim_complete()` has the same unguarded zero-count exit, and it
is reachable: the `NFS4ERR_DELAY` gates in EXCHANGE_ID and CREATE_SESSION that
would hold every client off until the load settles are conditioned on
`server.nfs4_drc`, which defaults to off, while the load itself runs for any
durable KV module. With the reply cache disabled, a client whose record has
already streamed in can re-establish and send its global RECLAIM_COMPLETE while
another client's record is still in flight, take `pending_reclaim` to zero and
close the window on that client. It gets the same guard: the record is retired,
but the zero count only ends the window once the load is READY.

### Notes

The two commits are independent and separately revertable.

Four tests are added to `test_nfs_persist`, which needs no VFS, event loop or
KV backend: `nfs_recovery_load()` with a NULL vfs leaves
`persistence_disabled` false (it keys off the module name being `memkv`), which
is enough to drive the recovery state machine directly. The last one drives the
RECLAIM_COMPLETE path with a populated `to_reclaim` set and asserts through
`nfs_recovery_open_check()`, i.e. as the second client's CLAIM_PREVIOUS sees
it; with the guard removed from `nfs_recovery_reclaim_complete()` alone it fails
at the first in-grace assertion.

Worth flagging separately: no model-based test can currently observe any of
this. The MBT harness never calls `chimera_server_config_set_kv_module()`, so
every quint run is `memkv`, `nfs_recovery_begin_grace` short-circuits, and the
server-wide grace window is never open in any of them; there is no reboot event
in the NFSv4 model either. Closing that gap wants a `kv_module` knob on
`struct mbt_env_opts` plus a restart action in the spec, which is a larger piece
of work than this fix.


arenaud16:
## Code review

Reviewed `ed8a0dd4..d027e8e5` (3 files, +230/-1) plus the surrounding code:
`nfs4_recovery.c`/`.h` in full, the hydration half of `nfs4_drc.c`, the gates in
`nfs4_proc_open.c`, `nfs4_proc_reclaim_complete.c`, `nfs4_proc_exchange_id.c`,
`nfs4_proc_create_session.c`, `nfs4_session.c`, `nfs4_lease.c`, `nfs.c` init,
`tests/CMakeLists.txt`, and the quint harness plus `ext/specs`.

Both diagnoses are correct and both fixes are minimal and correctly ordered. One
blocker: the third claim in the description -- that the identical zero-count exit
in `nfs_recovery_reclaim_complete()` is unreachable -- does not hold in the
default persistent configuration.

### Strengths

- Both root causes are real, and I traced each independently.
  `nfs4_drc_ensure_client()` (`nfs4_drc.c:433`) callocs the restored record, and
  `chimera_nfs4_open()` (`nfs4_proc_open.c:1879-1883`) refuses `!is_reclaim &&
  !done` with no deadline, so the pre-fix behaviour really was a permanently
  unopenable mount. And `nfs_recovery_finalize_load()` (`nfs4_recovery.c:341-366`)
  really never re-arms `in_grace` -- it only ever clears it -- so a sweep that
  closes the window during the load is unrecoverable.
- The `loading` read is correctly placed and non-racy in both stale-read
  directions: `load_state` is monotonic to READY, and after READY the scan adds
  no more records, so a stale `false` cannot admit a premature close; a stale
  `true` costs at most one tick.
- `nfs_recovery_loading()` returning true for IDLE (not just RUNNING) means the
  guard also covers the pre-kickoff window.
- No on-disk impact: no serializer, deserializer or record layout is touched, and
  the restored `reclaim_complete` bit is derived rather than persisted. A mixed-
  version rolling restart reads and writes byte-identical records.
- No CMake change was needed and none was made -- `tests/CMakeLists.txt:47-65`
  already compiles `nfs4_recovery.c`, `nfs4_drc.c`, `nfs4_lease.c` and
  `nfs4_session.c` into `test_nfs_persist`, and the new includes are satisfied by
  that target.
- The comment explaining why the field is set directly rather than through
  `nfs4_client_mark_reclaim_complete()` (`nfs4_ct_lock` is not recursive) is
  correct -- `nfs4_session.c:867` takes that same lock.

### Critical

**The zero-count exit in `nfs_recovery_reclaim_complete()` is reachable, with the
same silent-lock-loss outcome this PR exists to prevent.**

`nfs4_recovery.c:621-633`, specifically the unguarded `pending_reclaim == 0 ->
in_grace = false` at :628-631.

The unreachability argument rests on EXCHANGE_ID and CREATE_SESSION answering
`NFS4ERR_DELAY` until the load settles. Both gates are conditioned on the 4.1
reply-cache knob, not on persistence:

- `nfs4_proc_exchange_id.c:122` -- `if (chimera_server_config_get_nfs4_drc(...) && nfs_recovery_loading(...))`
- `nfs4_proc_create_session.c:53` -- same conjunction.

`nfs4_drc` defaults to 0 (`server.c:431`), and the comment at `nfs.c:421` states
the asymmetry explicitly: recovery-record persistence is independent and always
on. `nfs_recovery_load()` stores `rec->nfs4_drc` but nothing in
`nfs4_recovery.c` ever reads it except the info log at :166 -- kickoff (:482),
persist (:275, :317) and `open_check` (:544) all key off `persistence_disabled`
alone. So in a `server.kv_module = <durable>` deployment with `nfs4_drc` off,
neither gate fires.

Failure scenario: server restarts with a durable KV and `nfs4_drc` off. Clients A
and B both have persisted recovery records. The first compound kicks off the
load; the record scan streams records one at a time (`nfs_recovery_scan_cb`,
`pending_reclaim++` per key). A returns first, is not delayed, runs EXCHANGE_ID +
CREATE_SESSION, and -- having no state to reclaim, the common case for an idle
mount -- sends its global RECLAIM_COMPLETE immediately. A's record has loaded;
B's has not. `pending_reclaim` goes 1 -> 0, this function clears `in_grace`, and
nothing re-arms it. B's record lands a moment later, B's CLAIM_PREVIOUS gets
`NFS4ERR_NO_GRACE`, and B silently loses its locks. Same defect, same outcome,
same race window as the sweep bug the second commit fixes.

Fix is the same shape:

```c
if (rec->pending_reclaim == 0 && !nfs_recovery_loading(rec)) {
    rec->in_grace     = false;
    rec->grace_end_ns = 0;
}
```

`nfs_recovery_loading()` takes no lock, so calling it under `rec->lock` is safe;
hoisting it above the lock as the sweep does works too. If you would rather not
widen the change, file it and amend the description -- the current text asserts
unreachability, and a future reader will trust it. Either way the EXCHANGE_ID /
CREATE_SESSION reasoning needs the `nfs4_drc`-conditional caveat spelled out: it
is the load-bearing premise and it is only true with the reply cache enabled.

### Important

**1. The first commit does touch the server-wide window: a hydrated client's
`to_reclaim` record can no longer be retired.**

`nfs4_drc.c:452` interacting with `nfs4_proc_reclaim_complete.c:31-37` and
:46-50. The RECLAIM_COMPLETE handler calls `nfs4_client_mark_reclaim_complete()`
*first* and returns `NFS4ERR_COMPLETE_ALREADY` on a second global call, before
reaching `nfs_recovery_reclaim_complete()` at :47. Since the commit pre-stamps
the hydrated client, any global RECLAIM_COMPLETE it does send now short-circuits
at :34 and the recovery record is never retired. Hydration itself does not retire
it either. So for a hydrated client `pending_reclaim` cannot reach zero by any
path, and the server-wide window always runs to `grace_end_ns` -- 180 s by
default (`nfs4_lease.h:28`), during which every non-reclaim OPEN, LOCK, READ,
WRITE and size-changing SETATTR is refused for all clients.

Calibration: by the PR's own argument such a client does not send
RECLAIM_COMPLETE at all, so in the realistic case the record was never retired
before the change either and the practical delta is small. But "the server-wide
reboot window is untouched" is no longer strictly true -- the early-exit path
becomes structurally unreachable whenever any client hydrates. Either retire the
record at hydration time (a hydrated client is already barred from CLAIM_PREVIOUS
by `nfs4_proc_open.c:1873`, so this is consistent), or leave it and say so in the
commit message and in the comment at `nfs4_drc.c:441-451`.

**2. A hydrated client can no longer reclaim at all -- worth confirming this is
intended.**

With the mark set, `nfs4_proc_open.c:1873` (`is_reclaim && done ->
NFS4ERR_NO_GRACE`) refuses CLAIM_PREVIOUS from a hydrated client, even though
`nfs_recovery_open_check()` would have admitted it (its record is in
`to_reclaim`, unreclaimed). This looks harmless in practice: after hydration a
Linux client sees no BADSESSION, and if it later re-runs EXCHANGE_ID,
`nfs4_client_exchange_id()` case 2 (`nfs4_session.c:431-439`) returns
`CONFIRMED_R` on the same clientid, which drives nograce (CLAIM_NULL) recovery
rather than CLAIM_PREVIOUS. Still worth a line in the comment: during grace,
another client's reclaim of a conflicting share can win a file the hydrated
client legitimately held and can no longer reclaim.

**3. Test gap: none of the three tests exercises a gate function.**

`test_cross_reboot_reclaim_complete` asserts `nfs4_client_reclaim_complete()` is
true, not that an OPEN is admitted. That is a fair limit for the per-client gate,
since `chimera_nfs4_open` is not linked into this binary -- but
`nfs_recovery_open_check()` *is*: `nfs4_recovery.c` is compiled straight into
`test_nfs_persist`, and the function needs only a `struct nfs_client` with
`owner_string`/`owner_len`. A test that builds a `to_reclaim` entry, closes the
window the buggy way, and asserts `nfs_recovery_open_check(..., is_reclaim=true)
== NFS4ERR_NO_GRACE` before the fix and `NFS4_OK` after would pin the
client-visible symptom rather than the mechanism. That is also the natural home
for a regression test of the Critical item above.

**4. `test_grace_survives_sweep_during_load` never populates `to_reclaim`.**

`test_nfs_persist.c:498-542` only ever runs with `pending_reclaim == 0`. The
positive case -- records loaded, `pending_reclaim > 0`, READY, sweep must *not*
close -- is untested, as is the transition where the last reclaim retires and the
sweep then closes. One `to_reclaim` record would cover the state machine rather
than one branch of it.

### Minor

1. `test_nfs_persist.c:467-482` creates a duplicate-key session: part (2) creates
   a session with sessionid S, then calls `nfs4_drc_reconstruct_session()` with
   the same S. `nfs4_create_session()` (`nfs4_session.c:1285`) does no existence
   check, so the hash ends up with two entries under one key -- a state
   production cannot reach, since hydrate only runs on a lookup miss and is
   deduped by the hydra set. Teardown iterates in app order so nothing leaks, but
   the assertion at :483 is about the client, not the session, so the
   `nfs4_create_session` at :467 can simply be dropped.
2. `test_nfs_persist.c:452,466` pass `proto = 41`, but `nfs4_client_register()`
   ignores its `proto` argument entirely (`nfs4_session.c:146-220` hardcodes
   minor 0 into `nfs_client_alloc`). The "4.1 client" framing is nominal.
3. `test_nfs_persist.c:529-530` overstates what the deadline exit buys. "A load
   that never settles would hold the window open for ever" is true of
   `rec->in_grace` only; `nfs_recovery_open_check:544-547` forces in-grace
   behaviour while `load_state != READY` regardless, so a wedged load is already
   an indefinite `NFS4ERR_GRACE` outage. Preserving the deadline exit is still
   right; the stated reason is not the real one.
4. Comment style: the new comments write "RFC 8881 section 18.51.3" while every
   surrounding comment in `nfs4_recovery.c` (:556, :582, :598),
   `nfs4_proc_open.c` and `nfs4_proc_reclaim_complete.c` uses the section sign.
5. Micro: `nfs_recovery_sweep_once` now does an atomic acquire-load on every tick
   on every worker thread, including the common `!in_grace` case where it is
   discarded. Reading `load_state` after the `in_grace` check would skip it.
   Negligible either way.

### Claims in the description

1. Hydrated client never reclaim-complete; `SEQ4_STATUS_RESTART_RECLAIM_NEEDED`
   emitted nowhere; server-wide window untouched -- **holds, with one
   correction**. That status appears only as a constant in `nfs_common/nfs4.x:2691`;
   the only flags `nfs4_proc_sequence.c` sets are `RECALLABLE_STATE_REVOKED`
   (:44) and `CB_PATH_DOWN` (:58). `nfs4_drc_ensure_client` is reachable only
   from `nfs4_drc_reconstruct_session` -> `nfs4_drc_hydrate_session_cb`, so the
   mark cannot leak elsewhere. The correction is Important #1.
2. Grace can close before the load has counted; `finalize_load` never re-arms --
   **holds**. Also confirmed the blast radius is no wider than stated: while
   `load_state != READY`, `nfs_recovery_open_check` (:544-547) forces
   `in_grace = true` regardless of `rec->in_grace`, so a premature close only
   becomes visible after READY -- exactly when the loaded clients send
   CLAIM_PREVIOUS.
3. `nfs_recovery_reclaim_complete()`'s zero-count exit is unreachable --
   **does not hold**. See Critical.
4. The two commits are independent and separately revertable -- **holds**.
   Disjoint production files. In the shared test file, `34416948` is a single
   hunk at line 343 and adds no includes; `d027e8e5` adds the `pthread.h` /
   `stdatomic.h` / `nfs4_lease.h` includes its own test needs. Each compiles
   standalone; only the three adjacent `main()` call lines are a merge detail.
5. No model-based test can observe this -- **holds**.
   `chimera_server_config_set_kv_module()` has exactly one caller in the tree
   (`daemon.c:703`, from the JSON config); the quint harness
   (`nfs3_mbt_common.h:794-1017`) sets about thirty config knobs and never that
   one, so `config->kv_module` stays empty and `vfs.c:661` maps empty -> memkv ->
   `persistence_disabled`. The `ext/specs` NFSv4 model does model the client-reboot
   verifier case and the per-client `reclaimDone` gate (`nfs4_state.qnt:56`,
   `nfs4.qnt:697-740`) but has no server-restart/hydration event. Worth noting:
   the model does pin the behaviour the first commit must not break
   (`openBeforeReclaim` -> `NFS4ERR_GRACE` for a freshly-established 4.1 client),
   and the new test 2 guards exactly that.

### Suggestions beyond the fix

- Add a sentence to `nfs4_drc.c:441-451` recording why the mark is safe *today*:
  no open or lock state is persisted (`nfs_kv_keys.h:62-68` -- recovery, epoch,
  session, reply, nfs3 reply, nsm state, nsm monitor), so there is nothing a
  hydrated client's reclaim could rebuild that a post-grace CLAIM_NULL cannot.
  The day open/lock state is persisted, that line needs revisiting, and the
  trigger should be written down where the line is.
- Consider aligning the `NFS4ERR_DELAY` gates with `persistence_disabled` rather
  than `nfs4_drc` (`nfs4_proc_exchange_id.c:122`,
  `nfs4_proc_create_session.c:53`). Today the recovery load runs in
  configurations where the gates protecting it do not, which is the root of the
  Critical item and will keep producing bugs of this shape.

Two pre-existing hazards, out of scope here -- neither is introduced or worsened
by this diff, and neither was verified beyond reading:

- `nfs4_ct_next_client_id` resets to 1 on restart (`nfs4_session.c:90`) and the
  cold-start scan does not reserve persisted clientids -- only
  `nfs4_drc_ensure_client:477-483` advances the allocator, and only when it
  actually restores. A new client registering before a hydration can be issued a
  clientid a persisted session later claims; `ensure_client` would early-return
  on the id match and bind the reconstructed session to the wrong client.
- `nfs_recovery_scan_complete` (`nfs4_recovery.c:407-426`) discards `error_code`.
  A KV scan that fails outright is indistinguishable from "nothing was
  persisted", and `finalize_load` then drops the grace window entirely -- logged
  only as `grace skipped` at info level. Neither commit here introduces a silent
  failure; this one was already there.

### Verdict

Ready to merge with fixes. Both diagnoses are correct, both fixes are minimal and
correctly ordered, nothing about the persisted format or its interpretation
changes, and the new guard is non-racy in both stale-read directions. The blocker
is claim 3: the zero-count exit in `nfs_recovery_reclaim_complete()` is not
unreachable, because the `NFS4ERR_DELAY` gates it relies on are conditioned on
`server.nfs4_drc` (default off) while the recovery load is conditioned only on
the KV module being durable -- so as it stands the PR ships the same
silent-lock-loss race it sets out to close, in the default persistent
configuration. Fix that, and separately confirm that making the grace window's
early exit unreachable for hydrated clients is the intended trade.

arenaud16:
Pushed a new round (`9bc3b5c0`, `26cba37a`); both commits amended in place so
they stay independent and separately revertable. Item by item:

**Critical -- zero-count exit in `nfs_recovery_reclaim_complete()`.** Confirmed
reachable exactly as described: both `NFS4ERR_DELAY` gates are
`nfs4_drc && loading`, `nfs4_drc` defaults to 0, and nothing in
`nfs4_recovery.c` reads `rec->nfs4_drc` except the info log. Guarded the same
way as the sweep (`loading` read before the lock, same stale-read argument), the
second commit's message and the PR description now say so instead of claiming
unreachability, and the comment at the guard names the `nfs4_drc` conditional as
the reason the gates do not cover it. Regression test added
(`test_grace_survives_reclaim_complete_during_load`): with the guard removed
from the reclaim-complete path alone it aborts at the first in-grace assertion
while the sweep test still passes.

**Important 1 -- hydrated client's record can no longer be retired.** Holds.
Kept the mark and documented the consequence rather than retiring the record at
hydration: hydration is its own async KV scan and can complete before the
recovery scan has streamed that client's record in, so a retire-at-hydration
would need a placeholder record and a find-or-insert helper to avoid the same
ordering race this PR fixes twice already, and the practical delta is nil since
such a client never sends RECLAIM_COMPLETE. The comment at `nfs4_drc.c` and the
first commit's message now state it, along with the DESTROY_CLIENTID path that
does still retire it.

**Important 2 -- hydrated client can no longer CLAIM_PREVIOUS.** Intended;
now written down at the mark, including the conflicting-share point and the
reason it forfeits nothing today (no open or lock state is persisted; the
`CONFIRMED_R` path does not enter reboot reclaim). Also recorded the two
triggers for revisiting: persisting open/lock state, or starting to raise
`SEQ4_STATUS_RESTART_RECLAIM_NEEDED` for hydrated sessions.

**Important 3 and 4 -- test gaps.** The new test covers both: two `to_reclaim`
records streamed in the way `nfs_recovery_scan_cb` does, the OPEN-gate view of
each client (`NFS4_OK` / `NFS4ERR_GRACE` / `NFS4ERR_NO_GRACE` /
`NFS4ERR_RECLAIM_BAD`, plus `nfs_recovery_io_check`), the positive sweep case
(READY, one reclaim outstanding, tick must not close), and the last-reclaim
transition closing the window.

**Minor 1.** Dropped the live `nfs4_create_session()` in part (2); the
reconstruct now uses a sessionid the table does not know, which is the only
state production can reach.

**Minor 2.** Agreed the `proto = 41` framing is nominal; left as is since the
gate under test keys on the clientid, not the minor version.

**Minor 3.** Reworded: the deadline exit bounds `rec->in_grace` only, and
`nfs_recovery_open_check` answers `NFS4ERR_GRACE` until READY regardless.

**Minor 4.** Not taken: comments in this change stay ASCII (`section` rather
than the section sign), which is a deliberate constraint on my side.

**Minor 5.** Left as is; negligible, as noted, and moving the read under the
lock would only change the comment that explains it.

**Suggestion 1** is folded into the `nfs4_drc.c` comment. **Suggestion 2**
(gate EXCHANGE_ID / CREATE_SESSION on `persistence_disabled` instead of
`nfs4_drc`) is left for a separate change: it would add a retry round trip to
every client's first EXCHANGE_ID after a restart in configurations that never
hit either bug, and the targeted guard closes the hole without that.

The two pre-existing hazards (clientid counter restarting at 1 without
reserving persisted ids; `nfs_recovery_scan_complete` discarding
`error_code`) are real and out of scope here; worth filing both.

PR #1698, opened 2026-09-16
smb: store and emit native Windows SIDs in security descriptors

arenaud16:
## What this does

Completes native Windows SID support in the SMB server. The VFS half is already
in `main` -- `struct chimera_principal` carries a `chimera_sid`, and memfs,
cairn and diskfs all persist owner/group SID companions. Nothing populated them,
because the SMB server still reduced every wire SID to a uid/gid at the protocol
boundary and dropped any ACE it could not map.

The result today is that a backend which persists native SIDs has nothing native
to persist: a stored descriptor holds the algorithmic `S-1-5-88-*` modefromsid
form, readable only by something that can map it back, and an ACE naming a
principal the identity authority does not know is silently lost.

**SET_INFO(SECURITY)** now keeps what it was given:

* an owner or group SID that resolves travels on as the `va_owner_sid` /
  `va_group_sid` companion;
* an ACE SID that resolves keeps the SID on its `USER`/`GROUP` principal
  alongside the numeric id;
* an ACE SID still unmappable on the final pass is stored as an opaque
  `CHIMERA_PRINCIPAL_SID` rather than dropped, so the descriptor round-trips
  losslessly -- the way NTFS keeps an entry for a departed domain user;
* algorithmic and well-known SIDs stay numeric or special, since they re-derive
  on emit.

**QUERY_INFO(SECURITY)** prefers what was stored: owner/group fields use the
backend's companions, an ACE whose principal carries a SID is emitted from those
bytes, and `OWNER@`/`GROUP@` substitute the stored owner/group SID so the ACEs
agree with the descriptor's own owner and group fields. Only SID-less principals
consult the identity cache, and the uncached-principal count that decides
whether to park for resolution skips anything already carrying a SID -- so a
descriptor built from stored SIDs never waits on the resolver.

A stored descriptor is therefore correct without winbind, survives an outage, and
is readable by any other consumer of the on-disk data.

## Testing

`smb_sd_sid_test` drives the decoder and emitter directly, with and without an
in-process identity authority: lossless decode/emit/decode for an unmappable
SID, resolution keeping both identities, the parked-then-opaque passes, and the
emitter preferring stored SIDs over the cache.

Because both that and `vfs_acl_sid_test` are extended-tier while the merge queue
gates on the quick tier, the last commit adds ground truth to the SMB2 info
probe, which already runs in the quick tier: a DACL carrying an unmappable user,
a well-known SID and a DENY ACE for a second unmappable user must come back
verbatim with type, mask and order intact. Before this series two of those three
are silently dropped, the DENY among them. It also pins the boundaries -- a file
with nothing stored still reports the algorithmic owner and group, and an owner
SID that resolves to nothing is not adopted, since silently taking one would
detach the file's owner from every POSIX check.

Verified end to end on a live cluster against a real Active Directory domain,
with an out-of-tree backend that persists the companions. Setting an ACL by raw
SID (the client was not domain-joined, so the name never resolved client-side)
results in a stored on-disk descriptor holding the real `S-1-5-21-...` SID in
the ACE and in both the owner and group fields, read back with no identity
authority in the loop.

## Notes

The `chimera_sid` string/binary codec and the SDK version 2 bump this builds on
are already in `main`; this is the protocol-server half of the same work.


arenaud16:
Don't understand why the pipeline failed it does not seem to be related to this code change trying again

